PR #109, opened 2021-07-05
Fix production deployment: install firebase function dependencies too

silviot:
Should fix [this failure](https://github.com/silviot/EisBuk/runs/2991991681?check_suite_focus=true).


ikusteu:
It does work both ways, but this is to keep things consistent (at least until we migrate all to yarn), as mixing yarn and npm deps installation can cause problems.

ikusteu (approved):

PR #111, opened 2021-07-07
Fixes #91 : User can select a subset of slots from copied week

### LINKED ISSUE #91: Allow to select a subset of slots to copy and then paste
At the week level.

Paste should be possible also for non-empty weeks

fadwamahmoud:
selectable copypaste functionality:
- implemented two new actions for adding ond deleting from copyWeekSlots
- selecting a slot highlights and deselecting de-highlights
- slots can be pasted into non-empty weeks

### DIFF
--- a/eisbuk-admin/src/components/slots/SlotListByDay/SlotsDay.js
+++ b/eisbuk-admin/src/components/slots/SlotListByDay/SlotsDay.js
@@ -21,6 +21,7 @@ import {
   deleteSlotFromClipboard,
   addSlotToClipboard,
 } from "../../../store/actions/actions";
+import { calendarDaySelector } from "../../../store/selectors";
 import { shiftSlotsDay } from "../../../data/slotutils";
 import LuxonUtils from "@date-io/luxon";
 import { DateTime } from "luxon";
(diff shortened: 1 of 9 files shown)


ikusteu (approved):

eisbuk-admin/src/components/slots/SlotListByDay/SlotsDay.js
```diff
@@ -50,6 +55,9 @@ const SlotsDay = ({
   const copiedWeekSlots = copiedWeek.slots
     ? copiedWeek.slots.map((slot) => slot.id)
     : [];
+
+  const checkSelected = (id) => copiedWeekSlots.includes(id);
+  const canClickSlots = enableEdit && copiedWeekSlots.length > 0;
```
silviot, line 60: I believe we should check on one more condition here: that the week currently displayed is the one currently in the clipboard.

With the current logic the user might select slots from two different weeks, and get surprising results on paste.

Other than this, I like the idea that, after deselecting all slots, the user needs to press the copy button again to get into "select mode".

silviot (changes requested):
Great job!

I have a couple of minor requests for change, but I think in general this PR nails it.

One more minor comment: let's give descriptive titles to PRs, even if they include the ticket number in the title. It will improve the PR listing.

silviot (approved):

PR #131, opened 2021-07-23
Fix/remove hacks

ikusteu:


.github/workflows/storybook.yml
```diff
@@ -13,6 +13,21 @@ jobs:
         with:
           path: "**/node_modules"
           key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+      - name: Create .env file with credentials
```
silviot, line 16: Is this really necessary for the storybook too?

Ideally these value should have sensible defaults that work out of the box when in development but be overridable via an `.env` file.

silviot (changes requested):

.github/workflows/storybook.yml
ikusteu: This is the only way I've gotten it to work. We're using store `<Provider>` as a wrapper for storybook build. Our store in turn connects to the Firebase and I've checked yesterday (through trial and error) and learned that chromatic uses `production` as `NODE_ENV`. This is the reason why storybook was working locally and failed on chromatic deployment (and the focal point of my debugging yesterday). 
If you wish, I can open up another issue to investigate which of these config values is privy data and which can maybe be included in the codebase, and/or make some rewrites to the store to exclude firebase connection (if possible), but this is the only way this can work with current code.

ikusteu: Without `.env` credentials, the storybook build inside `Storybook` workflow returned empty stories (which chromatic didn't like)

ikusteu: You can reproduce that behavior locally by removing `.env.development.local` and trying to run `yarn storybook`. The storybook builds but all of the stories are empty. The same thing happens in workflow, only requiring production credentials.

silviot: I see. I believe we should have a working dev (and storybook) environment even with no production env file, but if it's not trivial it's better to work on it separately in the future.

silviot (approved):

fadwamahmoud (approved):

PR #135, opened 2021-07-27
Feature/shared types

ikusteu:


eisbuk-install.sh
```diff
@@ -0,0 +1,19 @@
+# install and build shared files package
+cd eisbuk-shared
```
silviot, line 2: Nitpicking: I think this file could benefit from a `set -ex` so that it shows what it does at each step and it fails with an error if any intermediate step fails. Currently the script will exit with the status of the last command.

silviot: Also, let's set the executable bit and add a shebang to use `/bin/sh`.

silviot (approved):
Very nice! I just have a small comment for the installation shell script, but it's a small thing.

eisbuk-install.sh
ikusteu: Yeah, I was thinking about that as well, just didn't know how exactly to implement it. Thanks ;)

eisbuk-admin/src/components/slots/SlotListByDay/SlotsDay.stories.tsx
```diff
@@ -49,7 +49,7 @@ const slots = {
     categories: [Category.Course],
     type: SlotType.OffIceGym,
     durations: Object.values(Duration),
-    notes: "Pista 2\nPotrebbe non svolgersi",
+    notes: "Pista 2\n Potrebbe non svolgersi",
```
silviot, line 52: Most nitpicky comment I ever made: it was better without the space: with this change the resulting string will have a line that starts with a space. Not worth changing though. Just mentioning in case I'm missing something.

silviot (approved):

PR #139, opened 2021-07-28
Fixes #125 : localizing dates

### LINKED ISSUE #125: Internationalization: Localize Dates

fadwamahmoud:
Dates are now also internationalized using i18next and luxon
Fixes #125


ikusteu:
Would it be possible to add the "st"/"nd"/"rd"/"th" after the day date in english (for instance `Friday 21st July` instead of `Friday 21 July`. I understand it's additional functionality, so if you think it would be too much, let's discuss it if it's necessary.

fadwamahmoud:
> Would it be possible to add the "st"/"nd"/"rd"/"th" after the day date in english (for instance `Friday 21st July` instead of `Friday 21 July`. I understand it's additional functionality, so if you think it would be too much work, let's discuss if it's necessary.

 @ikusteu  I think it's a good enhancement,  however, there's no straight forward way to implement this. Unfortunately, Luxon doesn't support ordinal suffixes and if we wanted this feature it would be a major work-around for such a minor enhancement (IMO). Tell me what you think.

ikusteu:
@fadwamahmoud Yeah, we should put a pin in it.

ikusteu (approved):

PR #150, opened 2021-08-06
Fix/customer view fixes

ikusteu:


eisbuk-admin/public/locales/en/translation.json
```diff
@@ -1,118 +1,113 @@
...
+  "Authorization": {
+    "NotAuthorized": "You are not authorized to login",
+    "AdminsOnly": "This space is reserved for administrators",
+    "LoggedInWith": "You are logged in with: ",
+    "TryAgain": "Log out and try again with a different user"
+  },
+  "CustomerArea": {
+    "book_ice": "Book Ice",
+    "book_of_ice": "Book Off-Ice",
+    "calendar": "Calendar"
```
ikusteu, line 60: Please note these: The enums (when they're compiled) are values for each `Enum.Key`. This is why the keys here are written the same as enum values of said "routes".

eisbuk-admin/src/components/slots/SlotListByDay/SlotsDay.tsx
```diff
@@ -135,7 +138,22 @@ const SlotsDay: React.FC<SlotsDayProps> = ({
 
   return (
     <>
-      {view.startsWith("slots") ? (
+      {view === CustomerRoute.Calendar ? (
+        <Grid className={classes.bookingsListContainer} container spacing={3}>
```
ikusteu, line 142: This is a bit simpler in my opinion: Since you only have three options and you want to check if its one of the two, it's clearer to check if it's the remaining one, and if it isn't, by logic, it's automatically one of the two.

eisbuk-admin/src/containers/SlotsPageContainer.tsx
```diff
@@ -55,25 +56,7 @@ const SlotsPageContainer: React.FC<Props> = ({ slots, children, ...props }) => {
...
-      ? Array(7)
-          .fill(null)
-          .map((_, i) => currentDate.plus({ days: i }).toISODate())
-      : Array(5)
-          .fill(null)
-          .map((_, i) => currentDate.plus({ week: i }).toISODate())
-          .filter((date) =>
-            DateTime.fromISO(date).hasSame(currentDate, "month")
-          );
-
```
ikusteu, line 76: Good job on elaborating through JSDoc. No objections there, I've only updated to start with Capital letter and wrote the possible scenarios as bullet points. The mistake here, however, is that it's really uncommon to use JSDoc on variables rather than functions. 
This is ok because it makes more sence to create a function here. Check below please.

eisbuk-admin/src/containers/SlotsPageContainer.tsx
```diff
@@ -184,4 +167,22 @@ const SlotsPageContainer: React.FC<Props> = ({ slots, children, ...props }) => {
...
+const getDatesToDisplay = (currentDate: DateTime, view?: CustomerRoute) =>
+  view === CustomerRoute.BookIce
+    ? Array(5)
+        .fill(null)
+        .map((_, i) => currentDate.plus({ week: i }).toISODate())
+        .filter((date) => DateTime.fromISO(date).hasSame(currentDate, "month"))
+    : Array(7)
+        .fill(null)
+        .map((_, i) => currentDate.plus({ days: i }).toISODate());
+
```
ikusteu, line 187: In my opinion, it's better to write the getter or calculating functions outside of the component. This is good idea as long as the function doesn't do some state change or interact with the hooks. So please, take a look at what I did here.
This both improves readability (by not overcluttering the component code) and has some added benefits because you don't need to memoize the function as it's constant outside of the component. 
*If you want more info on `React.FC` rendering and rerendering flow and memoization let me know as it's a lengthy subject.

eisbuk-admin/src/enums/routes.ts
```diff
@@ -5,6 +5,12 @@ export enum Routes {
   Debug = "/debug",
 }
 
+export enum CustomerRoute {
+  Calendar = "calendar",
+  BookIce = "book_ice",
+  BookOffIce = "book_off_ice",
+}
+
```
ikusteu, line 13: I've named these "routes" even though they're views currently but will be used a routes later when we implement that logic. This is also to keep things consistent across views, path routes (in the future) and translations.

eisbuk-admin/src/containers/SlotsPageContainer.tsx
fadwamahmoud: Actually I did have something almost exactly like this (that's why I wrote JSDocs for it), later I felt like a function was overkill cause it's basically one argument and it's a global prop so I removed the function but forgot to remove the docs.
I didn't take into consideration the fact that it would be better outside the component so that's why I felt like the function was unnecessary, but yeah I see the point.

eisbuk-admin/src/components/slots/SlotListByDay/SlotsDay.tsx
fadwamahmoud: Yeah, I know. That was my original reasoning but for some reason it broke the admin slots view...I'll merge this PR and check if it works though.

Issue #153, opened 2021-08-09
Fix Notification button (OK)


ikusteu:
We need to fix notification button being clickable. I've identified that the problem is the button is not on top and has a different overlay over it (therefore it cannot be clicked). This might just be a CSS problem, so we should fix that.

PR #159, opened 2021-08-12
fix: snackbars unclickable and have no spaces in between

fadwamahmoud:
Fixes: #153 
Turns out this is a bug in `Notistack` which is solved in version 1.0.10, however this update only works with MaterialUI version 5 which is currently unstable, so I fixed the bug overriding the MUI root class which is passed to the `SnackbarProvider` component.

### DIFF
--- a/eisbuk-admin/src/App.tsx
+++ b/eisbuk-admin/src/App.tsx
@@ -18,14 +18,6 @@ import Notifier from "@/components/Notifier";
 import { currentTheme } from "@/themes";
 import makeStyles from "@material-ui/core/styles/makeStyles";
 
-/**
- * The styles part is a, hopefully, temporary workaround to a
- * bug in notistack that causes snackbars to be unclickable.
- * Notistack v.1.0.10 fixes that bug however,
- * it doesn't work with materialUI v.5
- * The bug is caused by the root class not having the 'pointerEvents'
- * property set to 'all', 'root' class here overrides that.
- */
 const App: React.FC = () => {
   const classes = useStyles();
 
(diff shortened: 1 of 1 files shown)


silviot (dismissed):
Quite a rabbit hole! Nice find (and workaround).
It makes sense to also briefly mention in the code that `useStyles` and the custom class are needed to work around a bug (that someday might be fixed but it might not be obvious to the person reading the code).

eisbuk-admin/src/App.tsx
```diff
@@ -37,4 +40,22 @@ const App: React.FC = () => {
...
+    /**
+     * This is, hopefully, temporary workaround to a
+     * bug in notistack that causes snackbars to be unclickable.
+     * Notistack v.1.0.10 fixes that bug, however,
+     * it only works with materialUI v^5, which is in pre-release currently.
+     * The bug is caused by the root class not having the 'pointerEvents'
+     * property set to 'all', 'root' class here overrides that.
+     * @TODO upgade MUI, notistack and remove this when MUI v5 becomes the LTS
+     */
+    pointerEvents: "all",
```
ikusteu, line 57: It's always best to write comments above (my preference) or inline with the part we're refering to. 
*fixed: the fix only works with MUI v^5, add `@TODO` flag*

eisbuk-admin/src/App.tsx
```diff
@@ -37,4 +40,22 @@ const App: React.FC = () => {
   );
 };
 
+// ***** Region Styles ***** //
+const useStyles = makeStyles((theme) => ({
```
ikusteu, line 44: I've removed the Theme typing altogether as, at most we don't need it, so it's best to have the `makeStyles` type it implicitly.

ikusteu (approved):
Please, in the future avoid uppercase letters in branch names: common convention plus easier to type ;) especially with longer branch names.

PR #161, opened 2021-08-16
Enhancement/mark all categories off ice

fadwamahmoud:
Fixes: #93 
- When admin clicks `off-ice-gym` or `off-ice-dancing` all categories get checked and disabled.
- Clicking `Ice` after that only sets disabled to false then admin can uncheck as they want.


silviot (approved):
LGTM. Good job @ikusteu @fadwamahmoud !

PR #165, opened 2021-08-17
Remove the amIAdmin cloud function

silviot:
Firestore rules protect the data against unauthorized access.
The function amIAdmin was used by the UI to determine if the currently logged
in user is an admin of a certain organization.
Being a serverless function it runs in a short lived environment that might need launch at
request time. So often the function would take one second to execute.

Using a query to firestore we can achieve the same result (determine if the current user
is an admin) with more predictable performance.
It's also a simpler solution with less code.


ikusteu (approved):
I always like it when we're using less code and more readable form. However, this seems a bit out of best practices...keeping the list of admins in local store and running auth logic against local store: unsecure. You do the ruling here, if you're ok with that, the code definitely looks a bit nicer.

silviot:
@ikusteu I totally agree with the spirit, and the UI should not be trusted with security concerns. But that's not the case here: [the firestore rules](https://github.com/eisbuk/EisBuk/blob/5aa61efb23544c0df52256a1cd7a00119d5de7bd/eisbuk-admin/firestore.rules#L25) do the check server side to make sure the user can only access what they should. The code in the UI is there to make sure unauthorized users get a meaningful message instead of a stack trace.

I pushed a commit to clarify these points.

ikusteu (approved):
All clear!

PR #166, opened 2021-08-18
Feature/atomic slots

### LINKED ISSUE #157: Add tests for SlotCard
This should be an easy task. I would have done it while writing the component, but postponed in order to focus on more pressing matters.
Should create tests (using `react-testing-library`) for `SlotCard` and it's subcomponents redux action dispatching.

ikusteu:
Fixes #157 
I've finally completed the `SlotCard` and `DateNavigation` components as well as debugged the tests. 
There are stories for visual tests and all of the functionality is covered within jest/rtl unit tests.
I've also provided the thourough documentation for each component and think this is the way to go with each component we develop in the future.
_note: none of the components have been integrated but rather "raised in isolation" and will be integrated after this PR_


eisbuk-admin/package.json
```diff
@@ -43,18 +43,21 @@
...
+    "test:quicktest": "echo Running tests with no emulators support && yarn jest --watch src",
+    "test": "echo 'Running all tests; using emulators.' && yarn test:emulators",
     "deploy:eisbuk": "firebase deploy --project eisbuk --only hosting:eisbuk",
     "deploy:production": "firebase deploy --project eisbuk --only hosting:igorice; firebase deploy --project eisbuk --except hosting",
     "deploy:preview": "firebase --project eisbuk hosting:channel:deploy",
     "eject": "react-scripts eject",
     "storybook": "start-storybook -p 6006 -s public",
     "build-storybook": "build-storybook -s public",
-    "chromatic": "npx chromatic --project-token dcmb4kmorxe"
+    "chromatic": "npx chromatic --project-token dcmb4kmorxe",
```
silviot, line 59: It was actually bad practice to put this (now expired) project token here. Let's remove the `chromatic` script altogether! We won't be running it manyally.

eisbuk-admin/package.json
```diff
@@ -43,18 +43,21 @@
...
+    "test-temp": "yarn jest SlotCard --watch"
```
silviot, line 60: This is probably a testing leftover

eisbuk-admin/src/components/atoms/CopyPasteWeekButtons/CopyPasteWeekButtons.tsx
```diff
@@ -0,0 +1,67 @@
...
+import Badge from "@material-ui/core/Badge";
+
+import DeleteIcon from "@material-ui/icons/Delete";
+import FileCopyIcon from "@material-ui/icons/FileCopy";
+import AssignmentIcon from "@material-ui/icons/Assignment";
+
+const CopyPasteWeekButtons: React.FC = () => {
+  return (
+    <>
+      {/* {showWeekDeleteConfirm ? (
```
silviot, line 13: The commented out code throughout this file can probably be deleted: the original is still present in `eisbuk-admin/src/containers/SlotsPageContainer.tsx`.

eisbuk-admin/src/components/atoms/DateNavigation/DateNavigation.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+export default {
+  title: "Date Navigation",
+  component: DateNavigation,
+} as ComponentMeta<typeof DateNavigation>;
+
+// we're using a standardized date not to fallback to current date and produce false positives with chromatic
+const defaultDateISO = "2021-03-01";
+const defaultDate = DateTime.fromISO(defaultDateISO);
+
+// ***** Region Different Timeframe ***** //
```
silviot, line 17: What's your opinion on the following syntax? VSCode supports folding regions when it's used.

```javascript
// #region Foo
const some = "code that can be folded";
const together = "using a tick next to line 1"
// #endregion Foo
```

eisbuk-admin/src/components/atoms/DateNavigation/DateNavigation.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+// All of the stories of this region are default
+// with changing timefrema (jump) value
```
silviot, line 20: Typo `timefrema`.

silviot (changes requested):
I'll need more time to go over the rest of the PR. For now I have a few comments.

eisbuk-admin/src/components/atoms/DateNavigation/DateNavigation.stories.tsx
ikusteu: Perfect! :ok_hand:

ikusteu: I have fixed the syntax files touched here and will open up a ticket for other files

eisbuk-admin/src/components/atoms/CopyPasteWeekButtons/CopyPasteWeekButtons.tsx
ikusteu: I would leave it as is to indicate that this part is still under construction, I've just put it here so that I can use it in stories. I will finish/test it out in the next PR while creating the customer area views (and stories).

eisbuk-admin/src/components/atoms/DateNavigation/__tests__/DateNavigationOther.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    test("should pass toggle state to render function", async () => {
+      const toggleTrue = "Toggle: true";
+      const toggleFalse = "Toggle: false";
+      renderWithRouter(
+        <DateNavigation showToggle>
+          {({ toggleState }) => (toggleState ? toggleTrue : toggleFalse)}
+        </DateNavigation>
+      );
+      screen.getByText(toggleFalse);
+      // this isn't so straight forward, this "checkbox" is the toggle button
```
silviot, line 32: I think we should only test things easy to test in this context.

I'd like to add some cypress test that exercise the whole application.
Cypress can [also be used to consume stories and run tests on them](https://cmc.jazkarta.com/members/silvio-tomatis/member-purchase-receipt?key=2021:8:24:10:32:59:319622).

eisbuk-admin/src/components/atoms/SlotCard/DurationsSection.tsx
```diff
@@ -0,0 +1,132 @@
...
+ */
+const DurationsSection: React.FC<Props> = ({
+  subscribedDuration,
+  enableSubscription,
+  ...slotData
+}) => {
+  const { t } = useTranslation();
+  const dispatch = useDispatch();
+
+  const { secretKey } = useParams<{ secretKey?: string }>();
```
silviot, line 49: I think the current URL scheme that embeds the secret key in the path should be kept: currently we have `/clienti/88de50d5-2b63-4652-9c86-bd91c6fe2c2f` (we should change `clienti` to `customer`) and that looks better than `/customer?secret_key=88de50d5-2b63-4652-9c86-bd91c6fe2c2f`.

eisbuk-admin/src/components/atoms/DateNavigation/__tests__/DateNavigationOther.test.tsx
ikusteu: Yeah...I wouldn't normally do this, but since it's a render prop, this additional test doesn't compromise the internal structure of the component. I agree that this is testing of internal state, which isn't preferred by RTL approach.

Sweet on the cypress + stories.

Will remove this.

eisbuk-admin/src/components/atoms/SlotCard/DurationsSection.tsx
ikusteu: Yes...first off, I completely agree that query params are somewhat ugly and I prefer not to use them as well (if not needed).
Secondly, I think I understand your confusion here: 
If you see the current code, you'll see that we already use the exact same implementation.The beauty of `react-router` is that, other than (traditional) query params, it features "route params", so this routes `path` is passed like this `customers/:secret_key`. This way the `:secret_key` is automatically recognized as a route param. So the real route would still look like it already does.

On a slightly unrelated note: there's no rule as to where the route params should be, so if we, for instance, wanted to use a sub-path language routing we could do something like this `path="/:lang/customers/:secret_key", and `react-router` and `useParams()` hook would return an object with this interface: `{lang: string, secret_key: string}`

eisbuk-admin/src/components/atoms/DateNavigation/__tests__/DateNavigationOther.test.tsx
ikusteu: Hey, on second thought, I would like to keep the tests ensuring that `toggleState` gets passed to the render function as it's part of core functionality. 
The not-so-straightforward solution `getByRole()` is actually a perfectly accepted method of testing with RTL. The comment is there to provide the reader of the code with explanation as I (due to the weird bug) wasn't able to explicitly (and descriptively) test with `getByTestId()`.
If you wan't me to remove some of these tests, please specify which and I will do so.

ikusteu: Or maybe just remove the state change (`.click()`) action part, but leave the test?

silviot (approved):
Very nice structure overall! I left a couple of minor comments; LGTM.

eisbuk-admin/src/components/atoms/DateNavigation/__tests__/DateNavigationOther.test.tsx
silviot: Let's leave it as is for now. I just wanted to share my thoughts about how I believe this should be ideally done, but let's not take care of it right now.

eisbuk-admin/src/components/atoms/SlotCard/DurationsSection.tsx
silviot: Nice! I didn't know the same infrastructure can retrieve URL parameters and path components (the way I usually call "route parameters").

PR #174, opened 2021-08-28
Try and fix tests

ikusteu:


eisbuk-admin/src/components/atoms/AttendanceCard/__testData__/dummyData.ts
```diff
@@ -19,8 +23,11 @@ export const users = [
   },
 ];
 
+// here we're using storybook date as default date and making sure that the slots starts at 13:00
+const luxonDate = DateTime.fromISO(__storybookDate__).plus({ hours: 13 });
+
 export const dummySlot = {
-  date: luxonToFB(DateTime.fromMillis(1629198000000)),
+  date: luxonToFB(luxonDate),
```
ikusteu, line 30: You can use __storybookDate__ constant as default date for testing as well. This is better as it should always render a proper `DateTime`. 
Your test was failing on GH and passing locally because of timezone (I think) as milliseconds are absolute time passed from certain date (IDK which, probably the UNIX date) and luxon automatically sets that in the context of timezone. 
The GH test was two timezones begind us (the `screen.debug()` of GH tests showed `11:00 - 13:00` instead of `11:00 - 13:00`).

ikusteu: `__storybookDate__` is currently something random. I've changed it in my branch to "2021-03-01" for convenience (monday, 1st of the month), but this works for now as well.

fadwamahmoud (approved):

PR #176, opened 2021-08-30
Update all packages to latest versions

### LINKED ISSUE #115: Update packages to latest
This is actually realted to `firebase` and `firebase-tools` at the top level (client facing) app, Everything else is up to date, but updating one of those two causes 401/404 whe calling cloud functinos. We need to:

- investigate which of these packages is causing problems
- how and why the problem arises
- upon solving update everything to @latest version

silviot:
Fixes #115


silviot:
I should check wether `yarn upgrade` does what I expect it to. In this case it did not. Closing.

PR #178, opened 2021-08-30
attendanceCard UI

fadwamahmoud:


eisbuk-admin/src/components/atoms/AttendanceCard/__tests__/AttendanceCard.test.tsx
```diff
@@ -18,21 +19,27 @@ const mockDispatch = jest.fn();
...
+      () => spyT.mockClear();
     });
     test("should render props", () => {
       screen.getByText("13:00 - 15:00");
       screen.getByText("Saul");
       screen.getByText("Walter");
-      screen.getByText(Category.Competitive);
+      // screen.getByText(Category.Competitive);
+      expect(spyT).toHaveBeenCalledWith(`Categories.${Category.Competitive}`);
+      expect(spyT).toHaveBeenCalledWith(`SlotTypes.${SlotType.Ice}`);
```
ikusteu, line 37: Nice...so yeah, we're implementing this for now but will return to `screen.getBytext` when we implement `i18next` with tests.

ikusteu: So, yeah...maybe just leave a `@TODO` here.

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +66,86 @@ const AttendanceCard: React.FC<Props> = ({
...
+                <ListItemAvatar>
+                  <EisbukAvatar {...user} />
+                </ListItemAvatar>
+                <ListItemText primary={user.name} />
+                <ListItemSecondaryAction>
+                  {absenteeButtons}
+                </ListItemSecondaryAction>
+              </ListItem>
+            );
+          })}
```
ikusteu, line 116: So yeah, this is ugly (having a large render mapping function inside return statement (a really hard to read).

ikusteu: The solution in my opinion would be to just create a `UserAttendance` component and move all of the rendering logic (and onClick events and all) there and only pass the necessary props.
This would greatly shorten the return of the mapping function and result in much more readable code 
```typescript
// note: this is just an example
{userBookings.map((userBooking) => <UserAttendance {...appropriateProps} />)}
```

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +66,86 @@ const AttendanceCard: React.FC<Props> = ({
...
+
+const translateAndJoinTags = (categories: Category[], type: SlotType) => {
+  const translatedCategories = categories.map((category) =>
+    i18n.t(`Categories.${category}`)
+  );
+  const translatedType = i18n.t(`SlotTypes.${type}`);
+
+  return `${[...translatedCategories, translatedType].join(" ")}`;
+};
+// ***** End Region Local Utils ***** //
```
ikusteu, line 133: Do the regions like this 
```typescript
// #region something

// someCode...

// #endregion something
```
Silvio's shown me this, VSCode lets you contract and expand regions if you do it like this.

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +66,86 @@ const AttendanceCard: React.FC<Props> = ({
...
+              >
+                <ListItemAvatar>
+                  <EisbukAvatar {...user} />
+                </ListItemAvatar>
+                <ListItemText primary={user.name} />
+                <ListItemSecondaryAction>
+                  {absenteeButtons}
+                </ListItemSecondaryAction>
+              </ListItem>
+            );
```
ikusteu, line 115: You literally need to copy this part of the code in to new fuctional component `UserAttendance`, TypeScript will tell you which imports/props you're missing. Add those props into local `interface Props` and consume the components in here like this
```
{userBookings.map(userBooking => <UserAttendance {...userBooking} />)}
```

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +66,86 @@ const AttendanceCard: React.FC<Props> = ({
...
-            </button>
-          </div>
-        );
-      })}
-    </div>
+              secondary={translateAndJoinTags(categories, type)}
+            />
+          </ListItem>
+          {userBookings.map((user) => {
+            const isAbsent = absentees?.includes(user.customer_id) || false;
```
ikusteu, line 85: false is unnecessary here

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
fadwamahmoud: I know, that's why I titled the commit as unfinished; I wanted to leave `AttendanceCard` as it is until `UserAttendance` was finished (I just pushed the finished version)

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +55,65 @@ const AttendanceCard: React.FC<Props> = ({
...
-    </div>
+              secondary={translateAndJoinTags(categories, type)}
+            />
+          </ListItem>
+          {userBookings.map((user) => {
+            const isAbsent = absentees?.includes(user.customer_id);
+
+            return (
+              <UserAttendance
+                key={`${id}-${user.customer_id}`}
```
ikusteu, line 78: `user.customer_id` is unique, therefore, it's unnecessary to complicate with the booking (slot) id.

eisbuk-admin/src/components/atoms/AttendanceCard/__tests__/AttendanceCard.test.tsx
```diff
@@ -2,43 +2,30 @@ import React from "react";
...
-  describe("Test marking attendance functionality", () => {
-    beforeEach(() => {
-      render(<AttendanceCard {...dummySlot} />);
-    });
-    test("should dispatch markAttendance with correct args", () => {
-      screen.getByText("👎").click();
-      expect(mockDispatch).toHaveBeenCalledWith(
-        mockMarkAttImplementation("heisenberg", true)
-      );
+      // TODO: move back to getByText when i18next is implemented
```
ikusteu, line 25: Please leave the TODO like this: `/** @TODO */`. It's cool if you hava an addon painting the TODOs automatically, but not everyone looking at the code will have it, and this way is default by VSCode. Plus, by using this format, you can write any string with `@` sign, for instance `/** @TEMP @Ivan :  */`, etc. and GitHub (which has really simple code highlighting recognizes the `/** @Something */` and colors it differently.

eisbuk-admin/src/components/atoms/UserAttendance/UserAttendance.tsx
```diff
@@ -0,0 +1,56 @@
...
+import Button from "@material-ui/core/Button";
+import makeStyles from "@material-ui/core/styles/makeStyles";
+import { ETheme } from "@/themes";
+import { markAttendance } from "@/store/actions/attendanceOperations";
+import { useDispatch } from "react-redux";
+
+type UserBooking = BookingsMeta & Pick<Customer, "certificateExpiration">;
+
+interface Props {
+  isAbsent: boolean;
```
ikusteu, line 17: This should be `isAbsent?: boolean`. This way you won't have to assert non-null type in the calling of the component (like you did in the `AttendanceCard`) and IMO it's generally good to leave the `boolean` props optional in most cases (as undefined is falsy and can be easily be used instead of `false` and passing of the props when calling the component is easier).

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +55,65 @@ const AttendanceCard: React.FC<Props> = ({
...
+                isAbsent={isAbsent!}
+                userBooking={user}
+              ></UserAttendance>
```
ikusteu, line 81: UserAttendance wont receive any children so this should be written as `<UserAttendance {...props} />` rather than `<UserAttendance><UserAttendance />`
In more general sense, you can always use this with `jsx`, even with basic tags, even though HTML doesn't support this sometimes, i.e.
```jsx
<div />

// will get compiled to HTML as

<div></div>
```

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +55,65 @@ const AttendanceCard: React.FC<Props> = ({
...
-            </button>
-          </div>
-        );
-      })}
-    </div>
+              secondary={translateAndJoinTags(categories, type)}
+            />
+          </ListItem>
+          {userBookings.map((user) => {
+            const isAbsent = absentees?.includes(user.customer_id);
```
ikusteu, line 74: This can be written directly into the component i.e.
```jsx
<UserAttendance 
  {...otherProps} 
  isAbsent={absentees?.includes(user.customer_id)} 
/>
```
And then you can return jsx diractly form the mapping function as `() => <UserAttendance />` without the unnecessary function block `{}`

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +55,65 @@ const AttendanceCard: React.FC<Props> = ({
...
-        const isAbsent = absentees?.includes(user.customer_id) || false;
-        return (
-          <div key={user.customer_id}>
-            <div>{user.name}</div>
-            <button
-              type="button"
-              onClick={() =>
-                dispatch(markAttendance(user.customer_id, isAbsent))
+    <Container maxWidth="sm">
+      <List className={classes.root}>
```
ikusteu, line 61: This should be part of parent component, as this will contain multiple `AttendanceCard` components. You can check in `BookingsByDay` that inside these two wrappers there is a mapping function to render all attendance cards (we're only interested in what's inside that mapping function for this component).

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +55,65 @@ const AttendanceCard: React.FC<Props> = ({
...
+  );
+  const translatedType = i18n.t(`SlotTypes.${type}`);
+
+  return `${[...translatedCategories, translatedType].join(" ")}`;
+};
+// #endregion Local Utils
+
+// #region Styles
+const useStyles = makeStyles((theme: ETheme) => ({
+  root: {},
```
ikusteu, line 104: Empty = unnecessary

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +55,65 @@ const AttendanceCard: React.FC<Props> = ({
...
+  listHeader: {
+    backgroundColor: theme.palette.primary.light,
+  },
+  slotWrapper: {
+    marginBottom: theme.spacing(1.5),
+    borderWidth: 1,
+    borderStyle: "solid",
+    borderColor: theme.palette.primary.main,
+  },
```
ikusteu, line 113: To be semantically correct, maybe name this `cardWrapper` or better you simply `wrapper`.

eisbuk-admin/src/components/atoms/UserAttendance/__testData__/dummyData.ts
```diff
@@ -0,0 +1,9 @@
+import { Category } from "eisbuk-shared";
+
+export const Saul = {
+  name: "Saul",
+  surname: "Goodman",
+  certificateExpiration: "2001-01-01",
+  customer_id: "saul",
+  category: Category.PreCompetitive,
+};
```
ikusteu, line 9: I like this :+1: , however you should write tests against in the parent component in this case.

eisbuk-admin/src/components/atoms/UserAttendance/__tests__/UserAttendance.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+      screen.getByText("Saul");
+    });
+    test("should dispatch markAttendance with correct args", () => {
+      screen.getByText("👍").click();
+      expect(mockDispatch).toHaveBeenCalledWith(
+        mockMarkAttImplementation("saul", false)
+      );
+    });
+  });
+});
```
ikusteu, line 38: This is nicely written, however, I told you to keep this test as it was:
- it is unnecessary to test in isolation as it's a simple functionality and this component won't be rendered anywhere else
- when you test it in the context of the parent component, you're implicitly testing the proper passing of the values from `AttendanceCard` props to `UserAttendance` props.

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +55,65 @@ const AttendanceCard: React.FC<Props> = ({
...
+  absent: {
+    backgroundColor: theme.palette.absent || theme.palette.grey[500],
+  },
```
ikusteu, line 116: This seems obsolete since you moved that part into the new component.

eisbuk-admin/src/components/atoms/AttendanceCard/__tests__/AttendanceCard.test.tsx
silviot: TIL - Good to know! I guess this is to support jsdoc syntax.

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +51,53 @@ const AttendanceCard: React.FC<Props> = ({
...
-              onClick={() =>
-                dispatch(markAttendance(user.customer_id, isAbsent))
-              }
-            >
-              {isAbsent ? "👎" : "👍"}
-            </button>
-          </div>
-        );
-      })}
+    <div key={id + "-" + timeString} className={classes.wrapper}>
```
ikusteu, line 56: Key is unnecessary here. The key is used by react-fabric (a reconciliation algorithm) to key react nodes in the component tree. It is used to:
- key nodes in the list (remember, you key the nodes, not the elements, i.e. you would do this:
 
```jsx
{map(() => 
  <AttendanceCard key="some_key">
    <Child />
  </AttendanceCard>
}
``` 
 not
```jsx
{map(() => 
  <AttendanceCard>
    <Child key="some_key" />
  </AttendanceCard>
}
``` 
as the `<div>` node you're keying is a child of `<AttendanceCard>` node when you put it all together.
- the other usage (since it is primarily used by react-fiber to recognize what needs to be rerendered), would be to intentionally force rerender of component by changing the key (which is really advanced usage and is rarely used)

eisbuk-admin/src/components/atoms/AttendanceCard/__tests__/AttendanceCard.test.tsx
```diff
@@ -1,38 +1,43 @@
...
+import { Category, SlotType } from "eisbuk-shared";
+import { customersSlot } from "../__testData__/dummyData";
+import i18n from "i18next";
 
-const mockMarkAttImplementation = (customerId: string, attended: boolean) => ({
+const mockDispatch = jest.fn();
+jest.mock("react-redux", () => ({
+  useDispatch: () => mockDispatch,
+}));
+const mockMarkAttImplementation = (customerId: string, attended?: boolean) => ({
```
ikusteu, line 13: Oh..the boolean here (and in the actual function should be boolean, not optional). The optional usage is "better" for components as it's often times used as a flag: `<ExpandableRegion isOpen  />`. Here the `isOpen` is optional boolean and if you put it like that, it is `true` and component will check for truthy of falsy (for rendering mostly), but in this case, we're storing the attended in a database which should always have a value. To ensure the optional `isAbsent` is one of `true` | `false` (not `null`, `undefined`, `0`, etc.) you can use `Boolean(isAbsent)` and it will cast `undefined` to `false`.

eisbuk-admin/src/store/actions/attendanceOperations.ts
```diff
@@ -9,5 +9,5 @@ import { Customer } from "eisbuk-shared";
  */
 export const markAttendance = (
   customerId: Customer["id"],
-  attended: boolean
+  attended?: boolean
```
ikusteu, line 12: The same boolean issue as in mock function

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +56,100 @@ const AttendanceCard: React.FC<Props> = ({
...
+    // presist attendance data to firestore
+    markAttendance && markAttendance(userBooking.customer_id, !isAbsent);
+  };
+
+  const checkLocalChange = (
+    bookingId: UserBooking["customer_id"],
+    isAbsentInFirestore: boolean
+  ): boolean =>
+    localAbsentees[bookingId] &&
+    localAbsentees[bookingId] !== isAbsentInFirestore;
```
ikusteu, line 94: All of this is unnecessary here. The local state should be kept local to `UserAttendance`. This way you don't pass too much data around, the state for `localAttendance` isn't needed anywhere outside the `UserAttendance` and in that case it would be a boolean(primitive) rather than an object(reference type), which is always better for state.
This part (`AttendanceCard`) should stay the same only it would make more sence to pass `attended` rather than `isAbsent`:
**AttendanceCard:**
```jsx
 <UserAttendance
      {...otherProps}
      attended={!absentees?.includes(user.customer_id)}
    />
```

**UserAttendance-logic:**
```typescript
// we're initializing the state with `attended` from props as they will be in sync initialy
  const [localAttended, setLocalAttended] = useState(attended);

  // no need for elaborate naming since it's the only onClick function and the code is small
  const handleClick = () => {
    const newAttended = !attended;
    setLocalAttended(newAttended);
    dispatch(markAttendance(customerId, true, bookingId)); // yes we should add booking Id to the markAttendance
  };
```
**UserAttendance-button:**
```jsx
<Button
      {...otherProps}
      colot={attended ? "secondary" : primary}
      onClick={handleClick}
    >
      {attended ? "👍" : "👎"}
    </Button>
```

ikusteu: Please see how simple it is: no `hasLocalChange`, no complications and passing of `toggle` function, etc.

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +56,100 @@ const AttendanceCard: React.FC<Props> = ({
...
+        />
+      </ListItem>
       {userBookings.map((user) => {
-        const isAbsent = absentees?.includes(user.customer_id) || false;
+        // set absence status with respect to firestore state
+        const isAbsent = Boolean(absentees?.includes(user.customer_id));
+
+        // check for discrepancy between local attendance (absence) status and the one in firestore
+        const hasLocalChange = checkLocalChange(user.customer_id, isAbsent);
+
```
ikusteu, line 115: unnecessary

eisbuk-admin/src/components/atoms/UserAttendance/UserAttendance.tsx
```diff
@@ -0,0 +1,64 @@
...
+  isAbsent?: boolean;
+  userBooking: UserBooking;
+  toggleAbsent: (UserBooking: UserBooking, isAbsent: boolean) => void;
+  hasLocalChange: boolean;
+}
+const UserAttendance: React.FC<Props> = ({
+  isAbsent,
+  userBooking,
+  toggleAbsent,
+  hasLocalChange,
```
ikusteu, line 26: What you need here is: `customerId`, `customerName`, `attended`, `bookingId`, nothing more, nothing less

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -36,28 +53,58 @@ const AttendanceCard: React.FC<Props> = ({
...
-            >
-              {isAbsent ? "👎" : "👍"}
-            </button>
-          </div>
+          <UserAttendance
+            key={user.customer_id}
+            slotId={id}
+            attended={!absentees?.includes(user.customer_id)}
+            userBooking={user}
+          />
```
ikusteu, line 78: {map((booking) => <UserBooking {...props} />} no need for `{}` block

eisbuk-admin/src/components/atoms/UserAttendance/UserAttendance.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const handleClick = (
+    slotId: Slot<"id">["id"],
+    userId: Customer["id"],
+    attended: boolean
+  ) => {
+    const newAttended = !attended;
+    setLocalAttended(newAttended);
+    dispatch(markAttendance({ slotId, userId, attended: newAttended }));
+  };
```
ikusteu, line 35: You can use closure here (no need to pass any params as all of the data is available in outer scope (component body). And then you would just pass `handleClick` as `onClick` on `<Button />` (without the need for `() => handleClick()`)

ikusteu (approved):

PR #183, opened 2021-09-06
Feature/slot form rehaul

### LINKED ISSUE #85: Slot form and data model rehaul
A slot should not have one datetime, but one date and multiple time intervals.
The form should allow the user to specify many intervals. For each interval the form will present these fields to fill in:
* start time
* end time
* duration to consider for bookkeeping (one, one and a half or two) is inferred from the duration of the `end -start` period
 The edit form should allow to edit the periods.

- [x] Create input functionality for said changes: - UI - SlotForm
- input functionality (actions)
- [x] UI: Edit small details in SlotForm:
- Differentiate "Create Slot" and "Edit Slot"
- [x] create validations for start and end time

ikusteu:
Fixes #85 
This is a complete SlotForm, with tests, storybook and full functionality (updated date model compatible). The component is currently in `@/temp/NewSlotForm` and is waiting for integration.


eisbuk-admin/src/components/temp/NewSlotForm/SelectCategories.tsx
```diff
@@ -0,0 +1,116 @@
...
+    // if slot type off_ice, disable all checkboxes and set value for "categories" to all
+    if ([SlotType.OffIceDancing, SlotType.OffIceGym].includes(slotType)) {
+      setValue(Object.values(Category));
+
+      setDisabled(true);
+    } else {
+      setDisabled(false);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [slotType]);
```
silviot, line 42: Shouldn't the dependencies array include all state dependent value used inside the clojure?

I expected this list to be `[slotType, setValue, setDisabled]`

eisbuk-admin/package.json
```diff
@@ -19,6 +19,7 @@
     "formik": "^2.1.5",
     "formik-material-ui": "^3.0.1",
     "i18next-http-backend": "^1.2.6",
+    "jest-environment-jsdom-sixteen": "^2.0.0",
```
silviot, line 22: This is already specified as a dev dependency, and is defnitely not needed at runtime.

eisbuk-admin/src/components/temp/NewSlotForm/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+export const slotToFormValues = (
+  slot: SlotInterface | undefined
+): (SlotFormValues & { date: DateTime }) | undefined => {
+  // fail early if no slot provided
+  if (!slot) return undefined;
+
+  const {
+    date: slotDate,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
silviot, line 22: Shouldn't we remove `id` from this assignment (since we actually don't use it) instead of disabling the linting warning?

silviot: I think I understand now. The `id` variable will be needed later on when coding the full functionality, so for now we like to keep it here.

eisbuk-admin/src/translations/en.json
```diff
@@ -130,5 +130,38 @@
...
+    "Ice": "ice",
+    "OffIceGym": "off-ice-gym"
+  },
+  "Category": {
+    "Course": "course",
+    "Competitive": "competitive",
+    "PreCompetitive": "pre-competitive",
+    "Adults": "adults"
+  },
+  "SlotForm": {
```
silviot, line 148: A key `SlotForm` is already present. We should not create another one.

eisbuk-admin/src/translations/it.json
```diff
@@ -129,5 +129,38 @@
...
+    "Ice": "ice",
+    "OffIceGym": "off-ice-gym"
+  },
+  "Category": {
+    "Course": "corso",
+    "Competitive": "agonismo",
+    "PreCompetitive": "pre-agonismo",
+    "Adults": "adulti"
+  },
+  "SlotForm": {
```
silviot, line 147: A key `SlotForm` is already present. We should not create another one.

eisbuk-admin/src/components/temp/NewSlotForm/utils.ts
ikusteu: No, to write it down (and make more sense then at the meeting): 

```typescript
interface InStruct {
  id: string;
  name: string;
  age: number;
}

interface OutStruct {
  name: string;
  age: number;
}

const ivan = {
  id: "13",
  name: "Ivan",
  age: 27,
};

const getPresonalInfo = (user: InStruct): OutStruct => {
  // split `id` of the user (type InStruct) and save the rest of the user as `presonalInfo`
  const { id, ...presonalInfo } = user;
  // return presonalInfo (type OutScruct)
  return presonalInfo;
};

const personalInfo = getPresonalInfo(ivan) // returns { name: "Ivan", age" 27 }
```

So here I'm using `id` declaration to put away the stuff I don't need from the object and return the rest (and the `id` is function scoped, so it gets garbage collected as soon as the function is done executing. This is just a syntactic sugar to not have to write it like this:

```typescript
const getPresonalInfo = (user: InStruct): OutStruct => ({
  name: user.name,
  age: user.age,
  // and you can imagine how many unnecessary lines
  // of code you get as the input object gets bigger
});
```

silviot: Totally makes sense. Thanks for the detailed explanation!

eisbuk-admin/src/translations/en.json
silviot: As discussed, I see why this is temporary.

eisbuk-admin/src/components/temp/NewSlotForm/SelectCategories.tsx
ikusteu: This is intentional as it was causing infinite loops.

silviot (approved):

PR #184, opened 2021-09-06
Feature/covid certificate

### LINKED ISSUE #170: Add field to the athletes form for EU Digital COVID Certificate
We need two new fields for the athlete form:
* _ Date of release of EU Digital COVID Certificate_ (Italian _Data di rilascio della Certificazione verde COVID-19_) A date field, indicating the start of the validity of the certificate
* _COVID-19 Certificate supsended_ (Italian _Certificazione verde sospesa_) A boolean (checkbox) that will be checked in case an athlete (even if vaccinated) came into contact with an infected person. It's not exactly clear yet how this will work, so for now we just let admins check/uncheck this.

### LINKED ISSUE #171: Show a warning mark on athletes lacking COVID-19 certification
To be implemented after #170.

Currently the athletes avatar sports a warning mark in case their medical certificate is expired.

We should have something similar (with a different symbol) for COVID-19 certification.
We should display a red mark if there is no certification, and a yellow mark if one is present but the _COVID-19 Certificate supsended_ flag is set to `true`.

ikusteu:
Fixes #170
Fixes #171 
I have created an entry in the `CustomerForm`, updated cloud function in `testData`, related to customers, updated `BookingsByDay` and `CustomerList` stories and `bookingDayInfoSelector`.
This is by no means an exhaustive TDD'd update, but rather make do to complete as fast as possible. Nevertheless, I think all the functionality works, but please review as I'm "cloudy" at best (since I've pulled an all nighter) so I might have missed something.

_note: the badge for COVID certificate is `"primary"` or `"error"` color (just like the medical certificate). I understand it is a temporary solution (not yellow), but unfortunately, Material-UI is pretty strict (limited) on which colors you can pass as prop to badge component._

### DIFF
--- a/eisbuk-admin/functions/src/testData.ts
+++ b/eisbuk-admin/functions/src/testData.ts
@@ -83,7 +83,7 @@ const createUsers = async (
         .plus({ days: _.random(-40, 200) })
         .toISODate(),
       covidCertificateReleaseDate: DateTime.local()
-        .plus({ days: _.random(-200, 40) })
+        .plus({ days: _.random(-500, 0) })
         .toISODate(),
       covidCertificateSuspended: _.sample([true, false])!,
     };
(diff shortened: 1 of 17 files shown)


silviot (approved):
I pushed a couple of changes. All the rest LGTM

PR #186, opened 2021-09-09
Feature/attendance 4

fadwamahmoud:


eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -54,6 +54,8 @@ const AttendanceCard: React.FC<Props> = ({
 
   const timeString = `${startTime} - ${endTime}`;
 
+  /** @TODO remove this when slots have intervals */
+  const dummyIntervals = ["13:00 - 14:00", "13:15 - 14:15"];
```
ikusteu, line 58: This should be received as props (so no need for `dummyIntervals`. If you need preview, that's exactly what storybook is for.

The props should look like this
```typescript
interface Props {
  ...propsYouHaveSoFar
  attendance: { // instead of absentees
    [customerId]: {
      bookedInterval: string | null
      attendedInterval: string | null  
    }
  }
  intervals: {
    [key: string]: {  // i.e. "08:00-09:30", don't ask why...we're doing it like this for now, but I'm rethinking it every day
      startTime: string // i.e. 08:00
      endTime: string // i.e. 09:30
    }
  }
}
```

eisbuk-admin/src/components/atoms/AttendanceCard/__tests__/AttendanceCard.test.tsx
```diff
@@ -54,10 +68,61 @@ describe("AttendanceCard", () => {
     });
     test("should disable attendance button while state and fb are synching", () => {
       screen.getByText("👎").click();
-      expect(screen.getByTestId("SaulGoodman")).toHaveProperty(
+      expect(screen.getByTestId("Saul-Goodman")).toHaveProperty(
```
ikusteu, line 71: please. please use snake-case for testids

eisbuk-admin/src/components/atoms/AttendanceCard/__tests__/AttendanceCard.test.tsx
```diff
@@ -54,10 +68,61 @@ describe("AttendanceCard", () => {
...
         "disabled",
         true
       );
     });
+    /** @TODO find a good test for options */
+    test("should render intervals in dropdown when athlete is present", () => {
+      screen.getByText("👎").click();
+      expect(screen.getByTestId("Saul-Goodman-input")).toBeInTheDocument();
```
ikusteu, line 79: There's no need for explicit assertion here `.toBeInTheDocument()` as `getByTestId()` will throw an error if test id isn't found within the document.

eisbuk-admin/src/components/atoms/AttendanceCard/__tests__/AttendanceCard.test.tsx
```diff
@@ -54,10 +68,61 @@ describe("AttendanceCard", () => {
...
+
+      // expect(screen.getAllByTestId("SaulGoodman13:00 - 14:00").length).toEqual(
+      //   customersSlot.intervals.length
+      // );
+    });
+    test("should render bookedInterval value as display value", () => {
```
ikusteu, line 85: `"should use bookedInterval as initial interval value"` seems a bit more accurate

eisbuk-admin/src/components/atoms/AttendanceCard/__tests__/AttendanceCard.test.tsx
```diff
@@ -54,10 +68,61 @@ describe("AttendanceCard", () => {
...
+      screen.getByText("👎").click();
+      expect(screen.getByTestId("Saul-Goodman-select")).toBeInTheDocument();
```
ikusteu, line 87: Explicit unnecessary

eisbuk-admin/src/components/atoms/AttendanceCard/__tests__/AttendanceCard.test.tsx
```diff
@@ -54,10 +68,61 @@ describe("AttendanceCard", () => {
...
+      const firstOption = screen.getByTestId("Saul-Goodman-select").firstChild;
+      expect(firstOption).toHaveTextContent("13:00 - 14:00");
+    });
+    test("should allow for selecting multiple intervals", () => {
+      fireEvent.change(screen.getByTestId("Saul-Goodman-input"), {
+        target: { value: "13:15 - 14:15" },
+      });
+      expect(screen.getByTestId("Saul-Goodman-input")).toHaveDisplayValue([
```
ikusteu, line 95: we're allowing customer to select only one interval, so ...this should be the opposite

eisbuk-admin/src/components/atoms/AttendanceCard/__tests__/AttendanceCard.test.tsx
```diff
@@ -54,10 +68,61 @@ describe("AttendanceCard", () => {
...
+      fireEvent.change(screen.getByTestId("Saul-Goodman-input"), {
+        target: { value: "13:15 - 14:15" },
+      });
+      fireEvent.change(screen.getByTestId("Saul-Goodman-input"), {
+        target: { value: "13:15 - 14:15" },
+      });
+      expect(screen.getByTestId("Saul-Goodman-input")).toHaveDisplayValue([
+        "13:00 - 14:00",
+      ]);
+    });
```
ikusteu, line 109: This become obsolete when the test/functionality is changed to select only one interval.

eisbuk-admin/src/components/atoms/AttendanceCard/__tests__/AttendanceCard.test.tsx
```diff
@@ -54,10 +68,61 @@ describe("AttendanceCard", () => {
...
+    // test("should disable dropdown when athlete is marked absent", () => {
+    //   // screen.getByText("👍").click();
+    //   expect(screen.getByTestId("WalterWhiteInput")).toHaveProperty(
+    //     "disabled",
+    //     true
+    //   );
+    // });
```
ikusteu, line 116: This seems like a valid test, but I understand this is a next TODO

eisbuk-admin/src/components/atoms/UserAttendance/UserAttendance.tsx
```diff
@@ -35,26 +50,60 @@ const UserAttendance: React.FC<Props> = ({ attended, userBooking, slotId }) => {
...
   );
+
   return (
     <ListItem className={listItemClass}>
       <ListItemAvatar>
         <EisbukAvatar {...userBooking} />
       </ListItemAvatar>
       <ListItemText primary={userBooking.name} />
-      <ListItemSecondaryAction>{absenteeButtons}</ListItemSecondaryAction>
+      <ListItemSecondaryAction>{absenteeButtons}</ListItemSecondaryAction>;
```
ikusteu, line 106: This will render `;` in the document

eisbuk-admin/src/store/actions/attendanceOperations.ts
```diff
@@ -14,3 +14,9 @@ export const markAttendance = (payload: {
   userId: Customer["id"];
   attended: boolean;
 }): void => {};
+
+export const selectInterval = (payload: {
```
ikusteu, line 18: disable the `eslint-no-unused-vars` for this line

eisbuk-admin/src/components/atoms/AttendanceCard/__tests__/AttendanceCard.test.tsx
fadwamahmoud: I use the first and last name props which have a capital first letter by default

eisbuk-admin/src/components/atoms/UserAttendance/UserAttendance.tsx
fadwamahmoud: this has been screaming warnings at me and I couldn't spot it at all...

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
fadwamahmoud: shouldn't `intervals` be a prop for `UserAttendance` ?
also `customerId ` is type string right?

ikusteu: You can use `SlotInterface` from `@/types/temp`. I've created that file to contain the new data model until it's intergrated.

ikusteu: So...you just need to extend `SlotInterface` instead of `Slot<"id">`

eisbuk-admin/src/components/atoms/AttendanceCard/__tests__/AttendanceCard.test.tsx
ikusteu: You don't need unique testids as you only need one customer for the testing. If you want to make sure that multiple customers render all the same (which is unnecessary in most cases), you can render multiple customers for that case.

ikusteu: The pure react-testing-library way would avoid testids altogether (and only test the stuff the user can see), but you can't really achieve TDD and full coverage that way. That's why I like to use testids, but that should be kept as simple as possible...and in most (99%) cases  you they should be explicitly set.

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -20,26 +20,25 @@ import { ETheme } from "@/themes";
...
 }) => {
   const classes = useStyles();
 
   // const { t } = useTranslation();
   // convert timestamp to luxon for easier processing
-  const luxonStart = fb2Luxon(date);
+  // const luxonStart = fb2Luxon(date);
 
   // convert durations to number values
   const durationNumbers = durations.map((duration) => Number(duration));
```
ikusteu, line 44: Yeah, I think you we're asking about this. Yes, this will absolutely go and get replaced by intervals.

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -54,13 +53,14 @@ const AttendanceCard: React.FC<Props> = ({
...
 
+
   return (
     <div className={classes.wrapper}>
       <ListItem className={classes.listHeader}>
         <ListItemText
           primary={
             <span>
-              {timeString} <b>({userBookings.length})</b>
+              {"13:00 - 14:00"} <b>({userBookings.length})</b>
```
ikusteu, line 63: You can calculate `startTime` as lowest `startTime` of all intervals and `endTime` as highest `endTime`of all intervals. Please avoid hardcoded values and pass everything from storybook.

eisbuk-admin/src/components/atoms/AttendanceCard/__testData__/dummyData.ts
```diff
@@ -39,6 +42,7 @@ export const emptySlot = {
 export const customersSlot = {
   date: luxonToFB(luxonDate),
   durations: [Duration["1.5h"], Duration["2h"]],
+  intervals: ["13:00 - 14:00", "13:15 - 14:15"],
```
ikusteu, line 45: Please simulate the data as it should be in the data model. Intervals should be 
```typescript
intervals: {[key]: {startTime: string, endTime: string}}
```

ikusteu: No whitespaces in the key

eisbuk-admin/src/components/atoms/AttendanceCard/__testData__/dummyData.ts
```diff
@@ -13,13 +13,15 @@ export const users = [
...
     category: Category.PreCompetitive,
+    bookedInterval: "13:00 - 14:00",
   },
   {
     name: "Walter",
     surname: "White",
     customer_id: "heisenberg",
     certificateExpiration: "2001-01-01",
     category: Category.PreCompetitive,
+    bookedInterval: "13:15 - 14:15",
```
ikusteu, line 24: no whitespace

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -70,8 +70,9 @@ const AttendanceCard: React.FC<Props> = ({
         <UserAttendance
           key={user.customer_id}
           slotId={id}
-          attended={!absentees?.includes(user.customer_id)}
+          attended={attendance?[user.customer_id].attended}
```
ikusteu, line 73: This is probably a TS error as `boolean` is expected and you're passing a `string | null` do the `Boolean()` assertion

PR #193, opened 2021-09-18
intervalCard tasks

fadwamahmoud:


eisbuk-admin/src/components/atoms/IntervalCard/IntervalCard.tsx
```diff
@@ -64,30 +65,13 @@ const BookingCard: React.FC<Props> = ({
...
-  //       </Typography>
-  //     </Typography>
-  //   </Box>
-  // );
 
   const handleClick = () => {
     /**
      * @FOR_FADWA_TASK_2
      * should call proper booking operation (with respect to `booked` flag)
      */
```
ikusteu, line 73: This is exactly the expected functionality, so you can remove the comment. Just a little cleaner way (since it's only one operation: you can remove the `{}` block and simply "return" `booked ? cancelBooking() : bookInterval()`. This will not return anything since both functions are `void`, just like the `handleClick` function, so type correct.

eisbuk-admin/src/components/atoms/IntervalCard/IntervalCard.tsx
```diff
@@ -64,30 +65,13 @@ const BookingCard: React.FC<Props> = ({
    * - you need to refactor this to use interval instead of calculating from duration
    * - should be straight forward, no fancy calculations or conversions (the start and end times are already received as `interval` properties as strings)
    */
```
ikusteu, line 67: This part is resolved, so you can remove this comment (`@FOR_FADWA`) as well as the below one (`TASK_2`)

eisbuk-admin/src/components/atoms/IntervalCard/__testData__/dummyData.ts
```diff
@@ -9,12 +9,23 @@ import { luxonToFB } from "@/utils/date";
...
+};
+export const nonBookedSlot: Props = {
+  date,
+  type: SlotType.Ice,
+  interval: {
+    startTime: "10:00",
+    endTime: "11:00",
+  },
+  notes: "",
+  booked: true,
```
ikusteu, line 30: There's no need copy paste all. In the tests you can just use:
```jsx
// for booked variant test
render(<IntervalCard {...baseProps} booked />)

// for not booked variant test
render(<IntervalCard {...baseProps} />)
```
Everything else is the same, except for interval property (which can be the same for both tests)

ikusteu: This is also the beauty of optional boolean flag we talked about on some of the previous PRs. In this case it's `booked` and the sole presence/absence of `booked` in the JSX will produce truthy/falsy value. The only scenario where you would need to explicitly set the `booked` property would be if you're using the condition, i.e.
```jsx
<IntervalCard {...someProps} booked={someCondition} />
```

eisbuk-admin/src/components/atoms/IntervalCard/__tests__/IntervalCard.test.tsx
```diff
@@ -1,16 +1,75 @@
...
+  ...jest.requireActual("react-redux"),
+  useDispatch: () => mockDispatch,
+  /** We're mocking this not to fail tests including SlotForm (this will be returning newSlotTime as null) @TODO remove in the future */
+  useSelector: () => null,
+}));
+
+jest.mock("react-router-dom", () => ({
+  ...jest.requireActual("react-router-dom"),
+  useParams: () => ({ secretKey: "secret_key" }),
+}));
```
ikusteu, line 24: You don't need this as you're not relying on the router anywhere within the component, nor do you need `secretKey` anywhere.

eisbuk-admin/src/components/atoms/IntervalCard/__tests__/IntervalCard.test.tsx
```diff
@@ -1,16 +1,75 @@
...
+import { SlotInterface, SlotInterval } from "@/types/temp";
+
+
+const mockDispatch = jest.fn();
+
+jest.mock("react-redux", () => ({
+  ...jest.requireActual("react-redux"),
+  useDispatch: () => mockDispatch,
+  /** We're mocking this not to fail tests including SlotForm (this will be returning newSlotTime as null) @TODO remove in the future */
+  useSelector: () => null,
```
ikusteu, line 18: If you were using selector inside of the component, always returning `null` wold produce errors in most cases. However you're not using `useSelector` within the component, so this is unnecessary as well.

eisbuk-admin/src/components/atoms/IntervalCard/__tests__/IntervalCard.test.tsx
```diff
@@ -1,16 +1,75 @@
...
+  ...jest.requireActual("react-router-dom"),
+  useParams: () => ({ secretKey: "secret_key" }),
+}));
+
+/** @TODO remove this when the i18next is instantiated with tests */
+jest.mock("i18next", () => ({
+  ...jest.requireActual("i18next"),
+  /** We're mocking this not to fail certain tests depending on this, but we're not testing the i18n values, so this is ok for now @TODO fix i18n in tests */
+  t: () => "",
+}));
```
ikusteu, line 31: You will actually need this. I would suggest to mock `t` implementation as `(label: string) => label`. This way, in test DOM, the appropriate value for i18n label will be rendered and you can use `screen.getByText(__bookInterval__)` for instance and have `const __bookInterval__ = "BookingCard.BookInterval"` saved in `@/lib/labels.ts` and in translation files have entries
```json
// en.json
"BookingCard": {
  "BookInterval": "BOOK INTERVAL"
}

// it.json
"BookingCard": {
  "BookInterval": "PRENOTA INTERVAL"
}
```
This way the testing is straightforward and you're reducing chances for typos.

eisbuk-admin/src/components/atoms/IntervalCard/__tests__/IntervalCard.test.tsx
```diff
@@ -1,16 +1,75 @@
...
+
+  // describe
+  test("should render properly", () => {
+    render(<IntervalCard {...bookedSlot} />);
+  });
   describe("Test button functionality ->", () => {
-    test("delete this test when you add your own", () => {
-      render(<IntervalCard {...baseProps} />);
+    const mockBookInterval = jest.spyOn(bookingActions, "bookInterval");
+    const mockCancelBooking = jest.spyOn(bookingActions, "cancelBooking");
```
ikusteu, line 45: Here you're spying on redux dispatches, while in our component we're receiving functions as props, so to test them out, you simply create two mock functions, pass them to the component and test them being called, i.e.
```jsx
const mockBookInterval = jest.fn()
const mockCancelBooking = jest.fn()

render(<IntervalCard booked bookInterval={mockBookInterval} cancelBooking={cancelBooking} />
/* some test actions */
expect(mockBookInterval).toHaveBeenCalled()
```

eisbuk-admin/src/components/atoms/IntervalCard/__tests__/IntervalCard.test.tsx
```diff
@@ -1,16 +1,75 @@
...
+
+
+const mockDispatch = jest.fn();
+
+jest.mock("react-redux", () => ({
+  ...jest.requireActual("react-redux"),
+  useDispatch: () => mockDispatch,
+  /** We're mocking this not to fail tests including SlotForm (this will be returning newSlotTime as null) @TODO remove in the future */
+  useSelector: () => null,
+}));
```
ikusteu, line 19: You won't need redux for unit tests of `IntervalCard` as it doesn't use store dispatching not selectors. All of the values, functions are passed through props.

eisbuk-admin/src/components/atoms/IntervalCard/__tests__/IntervalCard.test.tsx
```diff
@@ -1,16 +1,75 @@
...
+      render(<IntervalCard {...nonBookedSlot} />);
+      const nonBookedSlotLabel = `${nonBookedSlot.interval.startTime}-${nonBookedSlot.interval.endTime}`;
+      screen.getByText(nonBookedSlotLabel).click();
+      expect(mockDispatch).toHaveBeenCalledWith(mockBookInterval);
+    });
+    test("should fire `cancelBooking` on click, if the interval is booked", () => {
+      render(<IntervalCard {...bookedSlot}  />);
+      const bookedSlotLabel = `${bookedSlot.interval.startTime}-${bookedSlot.interval.endTime}`
+      screen.getByText(bookedSlotLabel).click();
+      expect(mockDispatch).toHaveBeenCalledWith(mockCancelBooking);
```
ikusteu, line 68: Thests are generally ok, if you were testing for redux dispatching, so just update the to use mock functions passed as props and find the label by text (like I've explained above with `i18n`), that way you would explicitly be testing function calls, but also implicitly, you would be testing that the proper button text is rendered (with respect to `booked` state).

eisbuk-admin/src/components/atoms/IntervalCard/IntervalCard.tsx
```diff
@@ -105,8 +76,13 @@ const BookingCard: React.FC<Props> = ({
           </Typography>
         </Box>
         <Box display="flex" flexGrow={1} flexDirection="column">
-          <Box display="flex" flexGrow={1} className={classes.topWrapper}>
-            {/* {timeSpan} */}
+          <Box
+            display="flex"
+            flexGrow={1}
+            className={classes.topWrapper}
+            onClick={handleClick}
```
ikusteu, line 83: We’re clicking on the button, not the card, so remove this. If you check the code down below, you’ll see that handler is already assigned to the button

eisbuk-admin/src/components/atoms/IntervalCard/IntervalCard.tsx
```diff
@@ -149,10 +124,10 @@ const BookingCard: React.FC<Props> = ({
               <Button
                 className={classes.actionButton}
                 onClick={handleClick}
-                color="secondary"
+                color={booked ? "secondary" : "primary"}
                 variant="contained"
               >
-                Cancel
+                {interval.startTime} - {interval.endTime}{" "}
```
ikusteu, line 130: This should be `BookingCard.CancelBooking` or `BookingCard.BookInterval`. But, for typos/testing consistancy, save it in a `@/lib/labels` as `const __cancelBooking__ = "BookingCard.CancelBooking"`, etc.

ikusteu: The time interval was ok as where it was. Check the storybook, the styles have been set up and shouldn't be changed (at least not for this PR).

ikusteu: Also the button color was fine as it was. For now.

eisbuk-admin/src/components/atoms/IntervalCard/__tests__/IntervalCard.test.tsx
```diff
@@ -1,16 +1,58 @@
...
+        customerId: Customer["id"];
+        bookedInterval: string;
+      }) => ({ params })
+    );
+    mockCancelBooking.mockImplementation(
+      (params: {
+        slotId: SlotInterface["id"];
+        customerId: Customer["id"];
+      }) => ({ params })
+    );
```
ikusteu, line 42: unnecessary

eisbuk-admin/src/components/atoms/IntervalCard/__tests__/IntervalCard.test.tsx
```diff
@@ -1,16 +1,58 @@
...
+    test("should fire `bookInterval` on click, if the interval is not already booked", () => {
+      render(<IntervalCard {...baseProps} bookInterval={mockBookInterval} />);
+      const nonBookedSlotLabel = `${baseProps.interval.startTime} - ${baseProps.interval.endTime}`;
+      screen.getByText(nonBookedSlotLabel).click();
+      expect(mockBookInterval).toHaveBeenCalled();
```
ikusteu, line 47: `screen.getByText(__bookInterval__).click()` where the `__bookInterval__` is a label in `@/lib/labels.ts`. `nonBookedSlotLabel` should be removed entirely. Repeat for the other test as well.

eisbuk-admin/src/components/atoms/IntervalCard/IntervalCard.tsx
```diff
@@ -105,8 +77,13 @@ const BookingCard: React.FC<Props> = ({
           </Typography>
         </Box>
         <Box display="flex" flexGrow={1} flexDirection="column">
-          <Box display="flex" flexGrow={1} className={classes.topWrapper}>
-            {/* {timeSpan} */}
+          <Box
+            display="flex"
+            flexGrow={1}
+            className={classes.topWrapper}
+            onClick={handleClick}
```
ikusteu, line 84: ...

ikusteu (approved):

PR #197, opened 2021-09-24
initial AttendanceSheet component

### LINKED ISSUE #172: Printable daily attendance
There should be a printable view that summarizes the day.

Each period from each slot should be listed with the name of each participant and space to allow a signature.

Something like this:

Attendance for 2021-08-26
===

| *10:10*-*10:40* |   |
|-------------|---|
| John Doe    |   |
| John Cena   |   |
| 10:10-11:30 |   |
| Walt Disney |  |
| Will Eiser  |  |
| Jane Doe    |  |

In the above example John Cena didn't show up.

fadwamahmoud:
fixes #172 
basic table which displays the date, intervals and customers who booked said interval


eisbuk-admin/src/components/layout/AttendanceSheet.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+  bookedInterval: intervals[1],
+  attendedInterval: null,
+};
+
+const customers = [
+  { ...saul, ...saulBookings },
+  { ...walt, ...waltBookings },
+];
+
+export const BareAppbar = (): JSX.Element => (
```
ikusteu, line 29: Semantics: This is more "Bare Table" than the "Appbar"

eisbuk-admin/src/components/layout/AttendanceSheet.tsx
```diff
@@ -0,0 +1,74 @@
...
+        {Object.keys(intervals).map((interval) => (
+          <TableBody>
+            <TableRow key={interval}>
+              <TableCell component="th" scope="row">
+                {interval}
+              </TableCell>
+            </TableRow>
+            {customers.map(
+              (customer) =>
+                customer.bookedInterval === interval && (
```
ikusteu, line 62: The customers shoul be shown based on `attendedInterval` as this will be the sheet for customers to sign (confirm) that they have attended the interval which will then bi billed to them (at the end of the month or whatever)

eisbuk-admin/src/components/layout/AttendanceSheet.tsx
```diff
@@ -0,0 +1,74 @@
...
+                  <TableRow key={123}>
```
ikusteu, line 63: The fixed key doesn't do you much good (ti's actually better to leave it empty, as react will then key them based on index, and show a warning). This way you're, sort of, letting the react fiber know that this doesn't need to be rerendered as it has the same key. *Sort of, because fiber will recognize this in some cases I think.

Maybe key them by `customerId` as you will have max one entry for customer per interval (list of jsx elements) and keys need to be unique only inside of rendered list of elements (doesn't matter if there is a same key on another element in another place of the component tree.

eisbuk-admin/src/components/layout/AttendanceSheet.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
```
ikusteu, line 1: For both files: This shouldn't be inside layout folder. For now you can put them in `atoms`, as we're mostly using that folder for new components...we will reorganize the folder structure later. 
The `layout` is for some layout components shared throughout the pages: admin bar, some shared headers (if we had any). Visually, you should think of layout as a frame for the content: so that the pages change the content inside the frame, while the frame stays the same (menu, logo, nav, etc.)

eisbuk-admin/src/components/layout/AttendanceSheet.tsx
```diff
@@ -0,0 +1,74 @@
+import * as React from "react";
```
ikusteu, line 1: `import React from "react"` is fine.

eisbuk-admin/src/components/layout/AttendanceSheet.tsx
```diff
@@ -0,0 +1,74 @@
...
+        <TableHead>
+          <TableRow>
+            <TableCell>
+              Attendance for {fb2Luxon(date).toFormat("EEEE dd-MM-y")}
+            </TableCell>
+          </TableRow>
+        </TableHead>
+        {Object.keys(intervals).map((interval) => (
+          <TableBody>
+            <TableRow key={interval}>
```
ikusteu, line 55: If you wanted to render `TableBody` for each interval (and you don't), you would key the `<TableBody>` component, not the `<TableRow>`.

_This is because, in the component tree, `TableBody` is a node within the component list (rendered by map function), with the children being child nodes of `TableBody` node,  and React Fiber only cares for keys of the list nodes (`TableBody` in this case)._

However, you don't want to render `<TableBody>` for each entry, but rather have:
```jsx
<TableBody>
  {intervals.map(() => 
    <TableRow key={interval}>
      ...
    </TableRow>
  }
</TableBody>
```   
**Things to keep in mind:**
- you only key the top level nodes in mapped list of components _(there is advanced use for keys, but no need to get into it now)_
- always check what you want to render as a list and what you want to render only once (as a list container)

eisbuk-admin/src/components/layout/AttendanceSheet.stories.tsx
ikusteu: And please put the `AttendanceSheet` in it's own folder, with `index.ts` and stories, like we're doing with all components.

eisbuk-admin/src/components/layout/AttendanceSheet.tsx
ikusteu: Right...now I see you're rendering multiple child nodes for each map function result. In those cases, it makes sense to use react fragment `<></>` to wrap all children: this is due to the rule that the mapping function should return only one component (only one fiber node) and in this case the fragment becomes that one node, and the components you're returning are the child nodes of that (fragment) node. 
Furthermore, since you want to key each node of the mapping function, here you should use `<React.Fragment key={interval} >` as the `React.Fragment` is the direct child of mapped list of components (`React.Fragment` is the same as `<>` but it accepts `key` prop unlike the short `<>` notation.

eisbuk-admin/src/components/layout/AttendanceSheet.tsx
```diff
@@ -0,0 +1,74 @@
...
+                customer.bookedInterval === interval && (
+                  <TableRow key={123}>
+                    <TableCell>{customer.name}</TableCell>
+                  </TableRow>
+                )
+            )}
+          </TableBody>
+        ))}
+      </Table>
+    </TableContainer>
```
ikusteu, line 71: @silviot advised on using native HTML + CSS rather than Material-UI, so you should take it up with him on this. Personally I condone this as, event though this will be less of a CSS practice, it is more efficient in terms of development.

eisbuk-admin/src/components/layout/AttendanceSheet.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { CustomerWithAttendance, SlotInterface } from "@/types/temp";
+
+export interface Props extends SlotInterface {
+  /**
+   * Record of customers who have booked (or are manually as attended),
+   * keyed by `customerId` and having all of the data for the customer plus
+   * values for `bookedInterval` and `attendedInterval`.
+   */
+  customers: CustomerWithAttendance[];
+}
```
ikusteu, line 19: Ok, so you're thinking about this a bit wrong: 
You're creating a component for the entire sheet and accepting props for a single slot: 
1. It would be good to develop the `AttendanceSheetSlot` first, accepting these props, but not rendering the sheet title (like the date) and render some info about the slot (as a title): it can be start/end time, slot type, etc.
2. Create an `AttendanceSheet` mapping through slots for a day and rendering `AttendanceSheetSlot` for each slot.

All in all the data flows between `AttendanceSheet` and `AttendanceSheetSlot` should be very similar (if not the same) as for the attendance page and `AttendanceCard` (without the functions of course)

eisbuk-admin/src/components/layout/AttendanceSheet.tsx
```diff
@@ -0,0 +1,74 @@
...
+  //   customers
+  // );
+
+  return (
+    <TableContainer component={Paper}>
+      <Table aria-label="simple table">
+        <TableHead>
+          <TableRow>
+            <TableCell>
+              Attendance for {fb2Luxon(date).toFormat("EEEE dd-MM-y")}
```
ikusteu, line 49: Cont'd on the comment about the `AttendanceSheetSlot`: this is the only thing you should change to slot info (explained above) and you can leave the rest as is.

ikusteu (changes requested):
All in all, you should fix the outlined errors and focus on developing this as an `AttendanceSheetSlot` and after that you can develop the entire sheet for a day (in my mind it would contain tables for each slot stacked on to of each other (visually forming one large sheet)

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <TableRow>
+          <TableCell>
+            {type} {notes}
+          </TableCell>
+        </TableRow>
+      </TableHead>
+      <TableBody>
+        {Object.keys(intervals).map((interval) => (
+          <React.Fragment>
+            <TableRow>
```
ikusteu, line 37: Missing key on `React.Fragment`

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,56 @@
...
+  intervals,
+  type,
+  notes,
+}) => {
+  return (
+    <Table aria-label="simple table">
+      <TableHead>
+        <TableRow>
+          <TableCell>
+            {type} {notes}
```
ikusteu, line 30: I would do it like this, so it's a bit prettier visually: `(${startTime} - ${endTime}) ${upperccaseType} ${notes}`

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheet.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import AttendanceSheetSlot from "./AttendanceSheetSlot";
+import {
+  baseProps,
+  saul,
+  walt,
+  intervalStrings as intervals,
+} from "@/components/atoms/AttendanceCard/__testData__/dummyData";
+export default {
+  title: "AttendanceSheetSlot",
```
ikusteu, line 10: You can name the titles of stories with whitespace: "Attendance Sheet Slot", this way it's a bit prettier in storybook.

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheet.tsx
```diff
@@ -0,0 +1,31 @@
...
+import AttendanceSheetSlot from "./AttendanceSheetSlot";
+
+interface SlotWithCustomers extends SlotInterface {
+  /**
+   * Record of customers who have booked (or are manually as attended),
+   * keyed by `customerId` and having all of the data for the customer plus
+   * values for `bookedInterval` and `attendedInterval`.
+   */
+  customers: CustomerWithAttendance[];
+}
```
ikusteu, line 14: Just leave a @TODO here to remove this when integrated, as this type exists in `@/types/components` in a branch not yet merged.

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { CustomerWithAttendance, SlotInterface } from "@/types/temp";
+
+export interface Props extends SlotInterface {
+  /**
+   * Record of customers who have booked (or are manually as attended),
+   * keyed by `customerId` and having all of the data for the customer plus
+   * values for `bookedInterval` and `attendedInterval`.
+   */
+  customers: CustomerWithAttendance[];
+}
```
ikusteu, line 16: @TODO +1

ikusteu, line 16: @TODO +1

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,56 @@
...
+                {interval}
+              </TableCell>
+            </TableRow>
+            {customers.map(
+              (customer) =>
+                customer.attendedInterval === interval && (
+                  <TableRow key={customer.id}>
+                    <TableCell>{customer.name}</TableCell>
+                  </TableRow>
+                )
```
ikusteu, line 48: Do the `${name} ${surname}`. In printed version, you should be able to differentiate between customers even if there are two customers with same name.

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheet.tsx
```diff
@@ -0,0 +1,31 @@
...
+export interface Props {
+  date: SlotInterface["date"];
```
ikusteu, line 16: The date, in this case, will be passed from redux store's `getCalendarDay`, so `DateTime`

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheet.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+};
+
+const customers = [
+  { ...saul, ...saulBookings },
+  { ...walt, ...waltBookings },
+];
+
+export const Default = (): JSX.Element => (
+  <AttendanceSheetSlot {...{ ...baseProps, customers }} />
+);
```
ikusteu, line 31: You should render the `AttendanceSheet` here so that you get the full context when looking at storybook.

ikusteu (changes requested):
Seems good, just these minor fixes. I would also use bold and thicker borders to separate the interval title and just a customer name.

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,94 @@
...
+        ))}
+      </TableBody>
+    </Table>
+  );
+};
+
+// #region Styles//
+const useStyles = makeStyles((theme) => ({
+  root: {
+    "& .MuiTableRow-root": {
```
fadwamahmoud, line 88: @ikusteu so this doesn't even show up in the inspector

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,94 @@
...
+  notes,
+}) => {
+  const classes = useStyles();
+
+  const overallInterval = getOverallInterval(Object.values(intervals));
+
+  return (
+    <Table aria-label="simple table">
+      <TableHead>
+        <TableRow className={classes.root}>
```
ikusteu, line 52: Be as semantic as possible with class names. I would name this `classes.slotHeading`

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,94 @@
...
+   * keyed by `customerId` and having all of the data for the customer plus
+   * values for `bookedInterval` and `attendedInterval`.
+   */
+  customers: CustomerWithAttendance[];
+}
+
+/** function that calculates start and end time for slot
+ * @param intervals - slot intervals array
+ * @returns overallInterval - { firstStartTime, lastEndTime}
+ */
```
ikusteu, line 27: Start the first line of doc as the second line of the comment (first line should be only `/**`, and then the rest from the next line)

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,94 @@
...
+  intervals: SlotInterval[]
+): {
+  overallStartTime: string;
+  overallEndTime: string;
+} => {
+  return {
+    overallStartTime: intervals[0].startTime,
+    overallEndTime: intervals[intervals.length - 1].endTime,
+  };
+};
```
ikusteu, line 38: I would name this `getTimeString` or `getSlotTimespan` and you can return the final string here (instead of start/end time object) so the in-component code is a tad cleaner.

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
ikusteu: I'm not sure what you're trying to achieve here, but here you're targeting all elements with the class `.MuiTableRow-root` nested within the element with the class `root`.The dynamic class name will vary, it's only `root` in js, but you get the point. So in this case (taking into account the jsx code), this way you're targeting an element with class `.MuiTableRow-root` within (nested) your `TableRow` component, which I'm pretty sure doesn't exist (didn't check the generated html). 
What I think you wan here is to target `<TableRow className={classes.root}>`. If that's the case, just lose the nested selector (`& .MuiTableRow-root`) and apply the styles to just the `root` class, in other words
```typescript
root: {
  borderWidth: 20 // in css-in-js, you can write pixels as number instead of string if the value is in px (but `20px` would work as well)
}
```

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,94 @@
...
+            )}
+          </React.Fragment>
+        ))}
+      </TableBody>
+    </Table>
+  );
+};
+
+// #region Styles//
+const useStyles = makeStyles((theme) => ({
```
ikusteu, line 86: Until you actually need the theme, for some classes, don't specify it as (unused) param (it's cleaner that way and eslint is happy)

ikusteu (commented):
There are some really small changes, as for the TS errors, I don't get any, nor do CI checks. 
What's happening for you locally (I think) is that you didn't run `./eisbuk-install.sh` or you simply need to restart the TS/ESLint server (`Ctrl + Shift + P)`

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheet.tsx
```diff
@@ -0,0 +1,32 @@
...
+export interface Props {
+  date: DateTime;
+}
+
+const AttendanceSheet: React.FC<Props> = ({ date, children }) => {
+  const { t } = useTranslation();
+  const classes = useStyles();
+  return (
+    <>
+      {" "}
```
ikusteu, line 17: The whitespace here is unnecessary

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,114 @@
...
+   * @TODO justify spaces in rows with 2 columns (name cell is much bigger than signature cell)
+   *
+   */
+  return (
+    <Table aria-label="simple table">
+      <TableHead className={classes.timeString}>
+        <TableRow>
+          <TableCell>
+            {timeString} {type.toUpperCase()}
+            {notes}
```
ikusteu, line 57: You can do a `<span>` around each string, it will be easier to style later.

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * function that calculates start and end time for slot
+ * @param intervals - slot intervals array
+ * @returns overallInterval - { firstStartTime, lastEndTime}
+ */
+const getTimeString = (intervals: SlotInterval[]): string => {
+  return `${intervals[0].startTime} - ${
+    intervals[intervals.length - 1].endTime
+  }`;
+};
```
ikusteu, line 33: This isn't right because you're depending on the intervals being sorted, whats more, even if the intervals are sorted, this might happen: `const sortedIntervals = ["09:00-10:00", "09:00-15:00", "10:00-11:00"]` and your result would be `"09:00 - 11:00"` when in fact end time should be `"15:00"`. 
The other thing...why not just take in intervals as they are (record insted of an array) and to the processing internally.

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,114 @@
...
+              </TableCell>
+            </TableRow>
+            {customers.map(
+              (customer) =>
+                customer.attendedInterval === interval && (
+                  <TableRow key={customer.id}>
+                    <TableCell className={classes.tableCell}>
+                      {customer.name} {customer.surname}
+                    </TableCell>
+                    <TableCell></TableCell>
```
ikusteu, line 80: I assume the empty `TableCell` will be place for a signature, you can do `<TableCell />` since it won't accept any children.

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,104 @@
...
+  /** UI todos:
+   * @TODO in page component render alternative background shades for slot table
+   * @TODO center each slot timestring
+   * @TODO justify spaces in rows with 2 columns (name cell is much bigger than signature cell)
+   *
+   */
+  return (
+    <Table aria-label="simple table">
+      <TableHead className={classes.timeString}>
+        <TableRow>
```
ikusteu, line 44: 1. Instead of assigning a `timeString` class to `TableHead` and targeting `.MuiTableRow-root` nested underneath, you can assign a `timeString` class to `TableRow`, which is the element you would be targeting with `timeString .MuiTableRow-root`. This way it's more straightforward/specific. It slightly optimizes CSS performance, which really isn't an issue here, but rather being cleaner with the code.
2. Maybe name this `slotHeading` or just `heading` since it covers the entire heading element rather than just the time string

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,104 @@
...
+          <TableCell>
+            {timeString} {type.toUpperCase()}
+            {notes}
+          </TableCell>
+        </TableRow>
+      </TableHead>
+      <TableBody className={classes.tableRow}>
+        {Object.keys(intervals).map((interval) => (
+          <React.Fragment key={interval}>
+            <TableRow>
```
ikusteu, line 54: The same as before: Rather assign a `tableRow` class to each `TableRow` than assigning it to a container and targeting nested table rows.

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,104 @@
...
+   * @TODO justify spaces in rows with 2 columns (name cell is much bigger than signature cell)
+   *
+   */
+  return (
+    <Table aria-label="simple table">
+      <TableHead className={classes.timeString}>
+        <TableRow>
+          <TableCell>
+            {timeString} {type.toUpperCase()}
+            {notes}
```
ikusteu, line 47: This should be bold, and I would also do `( ${timeString} )` adding parentheses to add some more visual distinction from intervals

ikusteu: Just an idea...maybe have the slot heading be bold, the interval heading be normal and customer names be a bit faded (the same color as table borders within customer names)

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,96 @@
...
+   * @TODO center each slot timestring (different width percentages for table cells doesn't work)
+   */
+  return (
+    <Table aria-label="simple table">
+      <TableHead>
+        <TableRow className={classes.slotHeading}>
+          <TableCell className={classes.bold}>
+            <span>{timeString} </span>
+            <span> {type.toUpperCase()}</span>
+            {notes && <span> ({notes})</span>}
```
ikusteu, line 47: Ok, so between these spans, I would create some margin. Also, maybe it would be good to use type badge (like in the slot component).

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheet.tsx
```diff
@@ -0,0 +1,32 @@
...
+  );
+};
+// #region Styles//
+const useStyles = makeStyles((theme) => ({
+  title: {
+    padding: 20,
+    background: theme.palette.primary.main,
+    fontFamily: theme.typography.fontFamily,
+    color: theme.palette.primary.contrastText,
+  },
```
ikusteu, line 29: Looks good with the contrast (greyscale view). I would just make the text bold, as it's a main title after all.

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheet.tsx
```diff
@@ -0,0 +1,32 @@
...
+export interface Props {
+  date: DateTime;
+}
+
+const AttendanceSheet: React.FC<Props> = ({ date, children }) => {
+  const { t } = useTranslation();
+  const classes = useStyles();
+  return (
+    <>
+      <div className={classes.title}>{t("AttendanceSheet.Date", { date })}</div>
```
ikusteu, line 17: Just a suggestion: I would maybe split the title and the date in two elements (`h3` and `span`) and put the date to the right edge of the row, without the parentheses (and possibly italic)

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,96 @@
...
+    borderBottom: "1px solid rgba(224, 224, 224, 1)",
+    paddingRight: 0,
+  },
+  bold: {
+    fontWeight: 600,
+  },
+  tableCell: {
+    borderRight: "1px solid rgba(224, 224, 224, 1)",
+    width: "50%",
+    color: theme.palette.grey[400],
```
ikusteu, line 92: Maybe make the color a tad less opaque (`grey[500]` or `grey[600]`)

fadwamahmoud: You mean more opaque? (I tried 500 and 600 and it makes it slightly darker)

ikusteu: I mean less opaque, so that it's visible on the printed paper. This might be too much.

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheet.tsx
```diff
@@ -0,0 +1,40 @@
...
+    background: theme.palette.primary.main,
+    fontFamily: theme.typography.fontFamily,
+    color: theme.palette.primary.contrastText,
+  },
+  date: {
+    position: "absolute",
+    right: "2rem",
+  }
+}));
+// #endregion Styles//
```
ikusteu, line 39: No need for `//` at the end, and also, write the region name as you would a var name: `PascalCase` for react components (in case you're making multiple components in the same file), `camelCase` for the rest: functions, styles, some random variables, etc.

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheetSlot.tsx
```diff
@@ -0,0 +1,154 @@
...
+                className={classes.type}
+                key="type"
+                color={slotLabel.color}
+              >
+                {t(slotTypeLabel[type])}
+              </Typography>
+            </Box>
+            {notes && <span className={classes.flexCenter}> ({notes})</span>}
+          </TableCell>
+        </TableRow>
```
ikusteu, line 80: I like the solution to show one time string on one end, and the rest on the other, but you can achieve that with way less markup. Have them all in one cell (the whole row), for that cell, you can use the `justify-content: "space-between"` and then have one `<Typography>` for timestring and one `div` containing type and notes.

This will also solve the problem of alignment: timestring is a bit indented with respect to interval strings.

ikusteu (changes requested):
I like the way the end result looks, I see some tasks to be done:
- [x] Fix formatting (some files aren't formatted)

**Create routing logic:**
- [x] Rename the page component to `attendance_printable` rather than `attendance` to avoid confusion as `root` (`/` route) is the main attendance page
- [x] Add `/attendance_printable` to `Route` enum (you can find it through `AppContent.tsx`)
- [x] Add new route and new page component to `AppContent` (so that we can access the page through the router (if you've never used `react-router`, you can just copy one of the existing routes and correct the path/component)

**Create UI routing to printable attendance:**
- [x] Add print icon to attendance page which will `Link` to `/attendance_printable`: the date should remain in redux all the same. You can add print button to attendance page in `DateNavigation` through `additionalButtons` prop

eisbuk-admin/src/AppContent.tsx
```diff
@@ -21,6 +21,7 @@ import CustomerAreaPage from "@/pages/customer_area";
 import useFirestoreSubscribe from "@/hooks/useFirestoreSubscribe";
 
 import { queryOrganizationStatus } from "@/store/actions/authOperations";
+import AttendanceSheet from "./components/atoms/AttendanceSheet/AttendanceSheet";
```
ikusteu, line 24: This is the cause of `/attendance_sheet` failing, you need to import a page component, not the component rendered on the page.

eisbuk-admin/src/AppContent.tsx
```diff
@@ -50,12 +51,14 @@ const AppContent: React.FC = () => {
...
       <PrivateRoute path={PrivateRoutes.Atleti} component={AthletesPage} />
       <PrivateRoute path={PrivateRoutes.Prenotazioni} component={SlotsPage} />
+
       <Route
         path={`${Routes.CustomerArea}/:secretKey/:customerRoute?`}
         component={CustomerAreaPage}
       />
       <Route path={Routes.Unauthorized} component={Unauthorized} exact />
       <Route path={Routes.Debug} children={<DebugPage />} />
+      <Route path={Routes.AttendancePrintable} component={AttendanceSheet} />
```
ikusteu, line 61: Good job, only use the different component (as mentioned above), also, this route should only be accessible to admin, so it also makes sense to put it above with the other private routes

eisbuk-admin/src/components/atoms/AttendanceSheet/AttendanceSheet.tsx
```diff
@@ -0,0 +1,46 @@
...
+    fontWeight: 600,
+    padding: 20,
+    background: theme.palette.primary.main,
+    fontFamily: theme.typography.fontFamily,
+    color: theme.palette.primary.contrastText,
+  },
+  date: {
+    position: "absolute",
+    right: "2rem",
+  },
```
ikusteu, line 43: Ok, so, first of good thinking on the absolute positioning, only the quirk with absolute positioning is: What is it absolute to? The answer is the first parent (up the tree) which is either `relative`, `absolute` or `fixed`. If no such parent is found, it falls back to the document.
So...I've added `position: "relative"` to a component parent to the date (title container)

eisbuk-admin/src/enums/routes.ts
```diff
@@ -3,6 +3,7 @@ export enum Routes {
   Unauthorized = "/unautorized",
   Debug = "/debug",
   CustomerArea = "/customer_area",
+  AttendancePrintable = "/attendance_printable",
```
ikusteu, line 6: :+1:

eisbuk-admin/src/pages/attendance_printable/index.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import { useSelector } from "react-redux";
+
+import Container from "@material-ui/core/Container";
+
+import AppbarAdmin from "@/components/layout/AppbarAdmin";
+import DateNavigation from "@/components/atoms/DateNavigation";
+import AttendanceSheet from "@/components/atoms/AttendanceSheet/AttendanceSheet";
+import AttendanceSheetSlot from "@/components/atoms/AttendanceSheet/AttendanceSheetSlot";
```
ikusteu, line 9: Ok...so this is where the `index.ts` file (in each component folder) is of use. What you wan't to do is: 
- Instead of importing `@/components/AttendanceSheet/AttendanceSheet` you want to import `@/components/AttendanceSheet` (index is automatic entry point of the folder, and the `default export` of `(...)/AttendanceSheet/index.ts` is the `AttendanceSheet` component.
- Also, instead of importing from `"@/components/atoms/AttendanceSheet/AttendanceSheetSlot"` you want to do `import {AttendanceSheetSlot} from "@/components/atoms/AttendanceSheet"` - since there can only be one default export, you simply do a named export in `(...)/AttendanceSheet/index.ts` file: 
```typescript
// ...
import AttendanceSheetSlot from "./AttendanceSheetSlot"
export { AttendnaceSheetSlot }
``` 
**As some rules of thumb:** 
1. Each file exported from the component folder should be exported from `index.ts` file, while the others are (sort of) private to that folder, this gives us a better overview of what is used locally and what is imported from elsewhere
2. In accordance to `1.`, you should only import from the folder (i.e.`@/components/AttendanceSheet`), and under the hood from `index.ts` file located in that folder, not dig through the folders files but rather, if you need some of the files within a folder, add the said file to `index.ts` file's exports.

ikusteu (changes requested):
Don't, for the life of you, edit anything in the printable sheet styling...I finally got it to work satisfactory (not the way I've imagined it, but close enough).

I wrote up the PR comments on where, what and why I've updated, so please take a note for future reference.

There are some additional stuff to take care of:
- [ ] Fix page `Route` component (comments provided)
- [ ] Fix `@/components/atoms/AttendanceSheet/index.ts` imports/exports (and apply to `@/pages/attendnace_printable/index.ts`
- [ ] Add print icon button to `DateNavigation` of attendance (`/`) page

eisbuk-admin/src/pages/attendance_printable/index.tsx
```diff
@@ -0,0 +1,38 @@
+import React from "react";
+import { useSelector } from "react-redux";
+
+import Container from "@material-ui/core/Container";
+
+import AppbarAdmin from "@/components/layout/AppbarAdmin";
+import DateNavigation from "@/components/atoms/DateNavigation";
+import AttendanceSheet from "@/components/atoms/AttendanceSheet";
+import { AttendanceSheetSlot } from "@/components/atoms/AttendanceSheet";
```
ikusteu, line 9: There should only be one import of one module (`import AttendanceSheet, {AttendanceSheetSlot} from "@/components/atoms/AttendanceSheet"`)

eisbuk-admin/src/components/atoms/AttendanceSheet/PrintButton.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export const PrintButton: React.FC<SlotButtonProps> = ({ size }) => {
+  return (
+    <>
+      <IconButton component={Link} to="/attendance_printable" size={size}>
+        <PrintIcon />
+      </IconButton>
+    </>
+  );
+};
```
ikusteu, line 25: This is just a copy/paste of `CopyButton` and most of the stuff is not neccessary, as the linter reports (a lot of unused imports). Normally I don't have anything against copying existing code and just tweaking the stuff you need (for efficiency), but make sure you completely adapt the new code (remove all of the stuff not necessary). 
Furthermore, there's no need for a separate component as this button should only be used in attendance page, so these three lines of code can easily be written in `@/pages/root/index.ts`.

Additional comment (with no relevance for the case at hand): if you were to use this as is, the react fragment is unnecessary since you're returning only one JSX.Element (IconButton) and everything else is nested inside.

eisbuk-admin/src/pages/slots/index.tsx
```diff
@@ -79,17 +81,20 @@ const SlotsPage: React.FC = () => {
...
+                    slotsToCopy={{
+                      day: Boolean(dayToPaste),
+                    }}
+                    {...{ date }}
+                  >
+                    <NewSlotButton />
+                    <CopyButton />
+                    <PasteButton />
+                  </SlotOperationButtons>
+                  <PrintButton />
```
ikusteu, line 96: This shouldn't be here

eisbuk-admin/src/AppContent.tsx
```diff
@@ -50,6 +51,11 @@ const AppContent: React.FC = () => {
       <PrivateRoute exact path={PrivateRoutes.Root} component={DashboardPage} />
       <PrivateRoute path={PrivateRoutes.Atleti} component={AthletesPage} />
       <PrivateRoute path={PrivateRoutes.Prenotazioni} component={SlotsPage} />
+      <PrivateRoute
+        path={Routes.AttendancePrintable}
+        component={AttendancePrintable}
+      />
+
```
ikusteu, line 58: :+1:

ikusteu (changes requested):
Most of the work is good, just some stuff I'd like you to correct:
- [x] Remove unnecessary `PrintButton` instances
- [ ] Delete the `PrintButton` component...it's unnecessary for it to be a component on it's own (three relevant lines of code)
- [x] Put the `PrintButton` in it's right place (`@/pages/root/index.tsx`)
- [ ] Fix up small issues mentioned in the comments (unnecessary stories, etc.)

To recap: print button should only be declared (and should only appear) in `@/pages/root/index.tsx` (attendance page), it should lead to `/attendance_printable` (as it is right now) so that the admin navigating through the app can easily access that feature (printable attendance).
_Also, the thing to keep in mind is that stories should only be updated if the component for which the stories were created was updated (by adding/removing/changing a feature)_

ikusteu (approved):

PR #198, opened 2021-09-26
Update/new data model

### LINKED ISSUE #71: Allow marking an athlete as present even when not booked
This should be a different state than booked and present.

The absentee key in the slots collection should change: instead we'll have an "athletes" key.

Current:
```
absentees : {
  "578a172d-8a59-4812-9a79-121a52bbd8e2": false
}
```
Wanted:
```
attendants: {
  "578a172d-8a59-4812-9a79-121a52bbd8e2": {
      // Booked and was absent
      duration_booked: "1.5",
      duration_actual: null,
      present: false,
  },
  "578a172d-8a59-4812-9a79-121a52bbd8e3": {
      // Booked and was present
      duration_booked: "1.5",
      duration_actual: "1.5",
      present: true,
  },
  "578a172d-8a59-4812-9a79-121a52bbd8e4": {
      // Did not book but showed up
      duration_booked: null,
      duration_actual: "1.5",
      present: true,
  },
  "578a172d-8a59-4812-9a79-121a52bbd812": {
      // Booked one hour and stayed two
      duration_booked: "1",
      duration_actual: "1.5",
      present: true,
  },
  "578a172d-8a59-4812-9a79-121a52bbd813": {
      // Booked one hour and stayed half an hour
      duration_booked: "1",
      duration_actual: "0.5",
      present: true,
  }
}
```

### LINKED ISSUE #72: Allow marking an athlete as having attended more than booked : 8 pts
Someone might book one hour but then stay for two. The app should allow the user to store this information.

Probably the "absent" button can become a menu with "mark absent"/"mark overtime" actions.

### LINKED ISSUE #148: Hard rewrite on the components
The components are way too cluttered with a lot of unnecessary data passed around. This would be quite some work. We should strive to reach as much of atomicity with the components as possible: components should be used as building blocks rather than huge inheritance monoliths and should be agnostic of environment as much as possible with only the necessary data passed in. This would make development, composition and visual/unit testing much easier and organized. The final idea would be to have each component in it's own folder containing:

- index.ts
- <ComponentName.tsx>
- <ComponentName.test.tsx> - a React testing library unit test
- <ComponentName.stories.tsx> - a visual representation of a component and used by chromatic for visual tests
- {...other} - local resources (for instance different background images imported in a `.ts` file and exported as array/key-value object

This is a strong **suggestion** and the priority of it plus the cost/benefit should still be defined and discussed.

- [x] Split components into reusable (environment agnostic) blocks
- [x] Write unit tests for (almost) each component 
- [x] Write stories for (almost) each component
- [x] Implement the app using said component and (as a side effect) stop prop drilling

### LINKED ISSUE #179: Create IntervalCard
We should create an `IntervalCard` component, used to show intervals (instead of slots) on `/customer_area` pages.
Shouldn't be too much work as most of the parts from `SlotCard` can be reused. It would be best to do it after #85 

## Step 1: Update `IntervalCard` (scaffolding provided) to have desired functionality
- [x] Rewrite time display to show start/end time from received interval
- [x] Add tests (and functionality) for button 

## Step 2: Update `IntervalCard`, (fix errors) and update/test `BookingCardGroup`
- [x] Revert time display to look like it did (UI), you can check `BookingCard` component - use storybook for this
- [x] Provide i18n labels for `BookingCard.bookInterval` and `BookingCard.cancelBooking` - use storybook to check this
- [x] Test and develop functionality for switching booked interval in `BookingIntervalGroup` (test scaffolding provided) - for now it should only be local state, without Redux dispatching

### LINKED ISSUE #180: Fix up useFirestoreConnect
When setting up `/customer_area` page views, I've implemented the date logic to route (rather than Redux store).
This, however produced some problems as the `useFirestoreConntect` for `slotsByDay` is called higher up the component tree (with respect to Redux date value), causing some of the date pagination not to work. 

The logic is stable for now: The default Redux date (initialized as `__storybookDate__`) has been fixed back to use `Date.now()` (if not within storybook environment). This results in querying for previous, current, and next month. This creates a convenience of two months as first slots will be added in September, and querying them in October will still produce results. However querying in November will already be out-of-time.

This should be done after the entire app has been moved to router dates and proposed solution would be to create a separate hook and make sure that no two instances of the same hook are rendered in the same time (on different levels of component tree).

### LINKED ISSUE #182: Integrate the new data model
This is a collection of data model integration tasks I've collected from all data model related tickets. I believe it's useful to have this a separate issue as:
- this should be done after we have all of the components/utils/selectors consuming the new data model
- it will serve as a checklist so that we don't leave anything out when integrating:

- Update cloud functions

- [x] update `testSlots`
- [x] create slot write -> attendance entry `dataTrigger`
- [x] update booking trigger (set initial attended value the same as booked value)
- [x] check `aggregateSlots`
- [x] remove `aggregateBookings`

- [x] create "migrations" to delete all slots, bookings and bookingsByDay so far

- Create Slot operations

- [x] create new copy/paste actions (populate existing functions with data)
- [x] create new slot operations (populate existing functions with data)

- Update components

- [x] update `DateNavigation` to use redux date exclusively
- [x] extract `SlotCard` to be only-admin slot with updated intervals (instead of durations)

- Update selectors

- [x] create selectors for `AttendancePage`
- [x] update `SlotsPage` selectors
- [x] update `CustomerAreaPage` selectors

- Update pages/routes

- [x] Create a page for attendance
- [x] update admin `/slots` page
- [x] Update `/customer_area` page with new `BookingCard` (when developed)
- [x] Update `/customer_area` routes (since we're using redux for date)

- Update types

- [x] Move all temp types to `eisbuk-shared` / local types
- [x] Keep deprecated types (for migrations and non integrated components)

ikusteu:
Fixes #182 
Fixes #71 
Fixes #72
Fixes #179 
Fixes #148 
Fixes #180 

Not to get into specifics, this PR solves the requirements detailed in solved Issues.

I would, however like to point out that there are still some minor fixes, cleans, updates that I will open up another issue for.

Some bugs/optimization requirements I've found off the bat:
1. The edit `SlotForm` is jumpy (closes on it's own) while the list gets populated with slots: I think the rerendereing of the page (for each new slot appearing) is what's causing it and it's not present when the test slots are created -> works as expected
2. The loading of the buttons is a bit slow (when switching toggle state on `/slots` page `DateNavigation`) -> this is entirely an profiler + optimization issue and should be solved later on as it works properly (only not as smooth)


silviot:
Running this locally with the emulators with an empty database it was not possible to create slots: no dates (and hence no `+` button) showed up.

Trying to book as an athlete I got the following exception:

```
Error: Actions must be plain objects. Instead, the actual type was: 'undefined'. You may need to add middleware to your store setup to handle dispatching other values, such as 'redux-thunk' to handle dispatching functions. See https://redux.js.org/tutorials/fundamentals/part-4-store#middleware and https://redux.js.org/tutorials/fundamentals/part-6-async-logic#using-the-redux-thunk-middleware for examples.
```

ikusteu:
Using this comment as a checklist for @silviot 's issues as well as providing some newly discovered bugs:
- [x] Solve empty `slotsByDay` -> empty week fallback
- [ ] Fix book slot bug (probably a dispatch of plain/non-thunk function)
- [ ] Fix persistance of deleted periond

**Newfound issues:**
- [ ] Slots get aggregated to a day before (due to time zone difference): Apply temp fix, and fix long-term by resolving Timestamp issue

ikusteu:
@silviot I've solved the issues you've found.
- the _no-days_ error was caused by the mistake of exiting early (within a selector) and returning empty object instead of an object of empty days
- the problem with booking operations was simple: they weren't created, I've overlooked them :embarassed: so, now I took care of that
- the intervals issue was a tricky one, the problem wasn't with the component, nor the thunk, but rather the `aggregateSlots` function as we're using `{merge: true}` flag while setting updated data, which causes `intervals` (as they're an object) to be merged (as part of recursive merge). I've solved this by pushing a `firestore.FieldValue.delete()` sentinel for each removed slot to make sure it gets deleted and still leave us an option to use `{merge: true}` (as we need it in order to place delete sentinel, for both intervals or an entire slot in case of delete)
- the issue I've found was caused by slots not aggregating properly because the `DateTime` was midnight (of UTC+1), and `Timestamp` is absolute (time diff from UNIX date), which cause the timestamp to be recognized as 11pm of the day before. This is an issue stemming from manually setting the "Timestamp", I've applied a patch-me-up solution and solving the Timestamp issue is my priority for tomorrow (along with european date and tests for `aggregateSlots`)

silviot:
@ikusteu I'm trying to deploy the functions to firebase but I get some generic errors in my terminal.

From the firebase console I got this:

```

migrateToNewDataModel
-
Provided module can't be loaded.
Did you list all required modules in the package.json dependencies?
Detailed stack trace: Error: Cannot find module 'eisbuk-shared/dist/deprecated'
Require stack:
- /workspace/dist/index.js
- /layers/google.nodejs.functions-framework/functions-framework/node_modules/@google-cloud/functions-framework/build/src/loader.js
- /layers/google.nodejs.functions-framework/functions-framework/node_modules/@google-cloud/functions-framework/build/src/index.js
 at Function.Module._resolveFilename (internal/modules/cjs/loader.js:889:15)
 at Function.Module._load (internal/modules/cjs/loader.js:745:27)
 at Module.require (internal/modules/cjs/loader.js:961:19)
 at require (internal/modules/cjs/helpers.js:92:18)
 at Object.458 (/workspace/dist/index.js:1:16791)
 at a (/workspace/dist/index.js:1:44100)
 at Object.837 (/workspace/dist/index.js:1:6703)
 at a (/workspace/dist/index.js:1:44100)
 at Object.341 (/workspace/dist/index.js:1:5195)
 at a (/workspace/dist/index.js:1:44100)
Could not load the function, shutting down.
```

It looks like the `eisbuk-shared/dist/deprecated.js` module doesn't make it to the firebase environement for some reason.

ikusteu:
Hmm…did you run eisbuk-install.sh, as it builds eisbuk-shared folder (we need to do it every time it gets updated as the build script spits out .js and .d.ts files which aren’t included in git). Idk, this is a first check in my debugging process…I can check it out tomorrow, I’m not home atm.

silviot:
@ikusteu After a first failure I retried running `eisbuk-install.sh` and a round of `yarn start` in `eisbuk-admin` but still faced the same problems.

Following [the advice here](https://github.com/firebase/firebase-tools/issues/968#issuecomment-460323113) I was able to make it work.

ikusteu:
Right...I'm reading up on CORS (as I think the time has finally came for me to understand it completely). Afterwards I'll google the exact issue, as, in my opinion it should be handled by Firebase out of the box (just a thought), so I'll check it out as well.

ikusteu:
@silviot I've fixed the issue with `eisbuk-shared/dist/deprecated`. The fix was extremely simple, it just took a little while to find it. The idea of bundling all of the functions in one file and then deploying was actually the way to go.
I've already implemented that to include `eisbuk-shared` in the webpack bundle, and not include the rest of the `node_modules` (to keep it light and I believe that the `firebase` installs third party libs everything from package.json). 
I didn't know that this didn't automatically apply to `eisbuk-shared/dist/deprecated`, so I manually added `eisbuk-shared/dist/deprecated` as included external modules (to be bundled with the code) and it works like a charm.

I also solved the CORS issue by inspecting the logs and realized I had some errors in the function, so firebase just short circuted that to CORS response for some reason. Any how I've added error handling and updated the tests a bit, so now the migration function works.

I have also created (and tested) `pruneSlotsByDay` migration to go through `slotsByDay` collection and delete empty days if any or delete the entire month's entry if all days are empty. The tests are passing, I just need to run it against staging firestore.

silviot:
A couple of fixes are needed:

* The storybook for some reason is failing after commit 06ca6f5ede685cb4abcfcbfbe5488b315e102cd3.
* I was not able to book a slot on https://eisbuk--198-merge-8f4ht7i1.web.app/customer_area/88de50d5-2b63-4652-9c86-bd91c6fe2c2f/book_ice

silviot (approved):

PR #201, opened 2021-10-01
Temp/new data model without tar

ikusteu:
This is a temp PR used to test out the migration function updates.


eisbuk-admin/functions/logs/1633031055859.json
```diff
@@ -0,0 +1,3 @@
+{
+  "lineReached": 204
+}
```
silviot: I believe this is one of the few cases when rewriting history is desirable.

It's also pretty convenient: you need to add `--amend` to your `git commit` message, and the changes will be incorporated in your last commit. You'll then need to `git push --force-with-lease` and alert other team members that pulled in your changes (TBF you're making their life a little miserable, but in this case the number of such colleagues was 0 as you were developing)

PR #220, opened 2021-10-28
Feature/attendance booked interval

### LINKED ISSUE #207: As an admin I should be able to tell if an athlete has over or underbooked
Currently the attendance view only shows one value for the booked interval.
Either the one the athlete originally selected, or the one the admin chose (if the admin marked over or under presence).

In cases where the athlete books an interval (say 9:00-10:00) and the admin changes it (say to 9:00-10:30) it should be clear from the UI. Maybe the orignal intervan could be shown on top or bottom of the current one.

ikusteu:
Fixes #207


silviot (approved):
I didn't manage to look into the code, but from a user's standpoint LGTM.

PR #221, opened 2021-10-28
Feature/separate lint job

ikusteu:
A small PR, adding a separate lint job to "Build and deploy preview" workflow (previously removed from build script, for convenience)


.github/workflows/build.yml
```diff
@@ -55,7 +55,7 @@ jobs:
         id: findcomment
         with:
           issue-number: ${{ github.event.pull_request.number }}
-          comment-author: 'github-actions[bot]'
+          comment-author: "github-actions[bot]"
```
silviot, line 58: No need to change this, but quotes are not needed here. Neither double nor single ones. You need quotes in YAML when you have a string like `"[1,2,3]"` that might represent a JSON object (YAML is a superset of JSON, so every JSON string is also a YAML string).

silviot (approved):

PR #225, opened 2021-10-30
Fix/customer area trap

### LINKED ISSUE #223: Solve customer area routing trap
Since we're using `react-router` for different views in `/customer_area` page, and the logic is set up so that the first view is pushed to history if the path to a route is not providede (if the route is `/customer_area`, it will automatically push `/customer_area/book_ice`).
This, however, introduces the bug of "trapping" the user on `/customer_area` page. For instance: if I load `/customer_area`, it will automatically push `/customer_area/book_ice` and I want to go back to before navigating to `/customer_area`, I will pres `<-` and will arrive to `/customer_area`, which will automatically push `/customer_area/book_ice` and so on.
The proposed solutions:
- Find a way to prune the `/customer_area` from browser/router history or perform `/customer_area` -> `/customer_area/book_ice` without pushing a new route, but rather mutate the latest route (sort of like `git commit --amend`)
- (sub-optimal) - add a timeout before redirecting so that the user can skip back if they're fast enough
- (sub-optimal) - remove the router from the picture and use the local state for view switching (loosing all of the benefits of using `react-router`)

ikusteu:
Fixes #223 
A quick fix, replace `history.push` with `history.replace`, thus removing "empty" route from history stack (and avoiding trapping the user when trying to navigate "back")


eisbuk-admin/src/pages/customer_area/CustomerNavigation/CustomerNavigation.tsx
```diff
@@ -67,7 +67,9 @@ const CustomerNavigation: React.FC = () => {
         __placeholder__,
         CustomerRoute.BookIce
       );
-      history.push(defaultRoute);
+      // here we're using `.replace` rather than `.push`
+      // to allow for back navigation (since it's an automatic redirect)
```
silviot, line 71: I would say we use `.replace` because it's semantically correct. The weird behaviour was a consequence of misusage of the history api.

Here we don't want to add something to the history. We want to effectively change the current history entry.

No need to change the code, but I believe it's a distinction worth mentioning.

eisbuk-admin/src/components/atoms/AttendanceCard/__tests__/AttendanceCard.test.tsx
```diff
@@ -76,7 +76,7 @@ jest.spyOn(i18n, "t").mockImplementation(((str: string) => str) as any);
 // aliases for thumbs button for easier access
 const thumbsUp = "👍";
 const thumbsDown = "👎";
-const trashCan = "🗑️";
+// const trashCan = "🗑️";
```
silviot, line 79: Unless there's a concrete plan for reintroducing code, let's remove it instead of commenting it out.

silviot (approved):

silviot:
I left a couple of minor comments

PR #229, opened 2021-11-02
Feature/functions esbuild

ikusteu:
I've created a build flow for `/functions`. The `watch` script is also there, but I had some issues integrating with test/dev environments. I believe the issue is with necessity of a more complex bash command (which I cannot seem to get right).

The current state of things is stable: 
- the build script IS slow, but that's due to pretty slow TS check before bundling (the script itself takes a little under a second and produces a lighter build compared to wabpack)
- all of the emulators (for local dev/test as well as CI) currently use the build script (as was the situation with webpack), however, I would like to have an option for hot module reloading (while the emulators are running), for which I have the script, only I cannot seem to get it right with `test`, `eumlators:inspect`, etc. scripts (purely a shell problem)

@silviot please take a look and decide if you wish try and find a solution tomorrow together (you being the shell master on these parts :wink: ) or merge if you wish (and add `watch` functionality later on) as current functionality is the same as the one we had with webpack


eisbuk-admin/package.json
```diff
@@ -43,11 +43,11 @@
     "start": "react-app-rewired start",
     "start:eisbuk": "REACT_APP_EISBUK_SITE=eisbuk.web.app NODE_ENV=production react-app-rewired start --reset-cache",
     "start:igorice": "REACT_APP_EISBUK_SITE=igorice.web.app NODE_ENV=production react-app-rewired start --reset-cache",
-    "emulators:inspect": "yarn --cwd functions run build && exec firebase emulators:start --project eisbuk --inspect-functions 9229",
-    "emulators:start": "yarn --cwd functions run build && exec firebase emulators:start --project eisbuk",
+    "emulators:inspect": "yarn --cwd functions functions:build && exec firebase emulators:start --project eisbuk --inspect-functions 9229",
```
silviot, line 46: Is this prefix needed? We're in the `functions` directory already, so I fail to see why we need to specify it once again.

ikusteu: No it is not...I thought about that as well...this is a bit more in detail, but yeah...

silviot:
@ikusteu [There are a couple of warnings](https://github.com/eisbuk/EisBuk/pull/229/commits/1c538130782d162b4af2b8b8d6560f2e0e3d33c4#diff-c8773c1cbca73f874ffdfdddda773df2e9258719c95c6944fe6fd0216d038c2eR6) that would make sense to resolve.
IIUC these lines
```javascript
import { __organization__ } from "@/lib/constants";
import { __functionsZone__ } from "@/lib/constants";
```
should become
```javascript
import { __functionsZone__,  __organization__ } from "@/lib/constants";
```

silviot (approved):

PR #236, opened 2021-11-11
local storage organization

fadwamahmoud:
- Organization is set in local storage during tests.
- getOrganization function checks for an organization value inside the local storage.
- it also allows for changing the organization during the lifetime of the app as opposed to exporting a constant.
- firestore instance is provided with `experimentalAutoDetectLongPolling` to enable cypress to handle its requests.


eisbuk-admin/src/components/DebugMenu.tsx
```diff
@@ -0,0 +1,72 @@
...
+        </MenuItem>
+        <MenuItem onClick={handleClose("createTestSlots")}>
+          Create slots
+        </MenuItem>
+      </Menu>
+    </>
+  );
+};
+
+export default DebugMenu;
```
ikusteu, line 72: I was going to ask: Why the two exact same `DebugMenu.tsx` files (this one and the one is `./layout`) and was going to suggest to use enums for function calls. 
Then I realized: the only difference between the two is usage of enums, so...why?

eisbuk-admin/src/utils/helpers.ts
```diff
@@ -103,3 +103,23 @@ can be published at https://appname--new-feature-randomhash.web.app/
 */
 export const getOrgFromLocation = (location: string): string =>
   location.replace(/--[^.]+/, "");
+
+export const getOrganization = (): string => {
+  return (
+    localStorage.getItem("organization") ||
+    getOrgFromLocation(window.location.hostname)
+  );
+};
```
ikusteu, line 112: I think that the `helpers` should be some function which process a given value (pure functions), not read from store, local store or somewhere else...for that I think that you can create a new file: `@/lib/getters.ts` and for now just put this function, but we can use it in the future for any other getters (if there will be any)

ikusteu (commented):
I've left some comments regarding the usage of `getOrganization` function. I understand that this is an a transitional solution and that the idea is to use redux to hold organization in the future. However, it is always a good idea to do things right the first way around and have the least amount of cleanup after.

ikusteu:
Also, the usage of `__organization__` unit tests was ok the way it was, as the unit tests don't depend on dynamic organization assignment.

eisbuk-admin/src/utils/helpers.ts
ikusteu: Also, here you can use `__organization__` as a fallback (as it handles additional case where the window is not defined)

eisbuk-admin/src/components/DebugMenu.tsx
fadwamahmoud: I think this is a mistake. It's from an older commit so I'm not exactly sure what happened.

eisbuk-admin/src/enums/store.ts
```diff
@@ -17,7 +17,7 @@ export enum Action {
   DeleteSlotFromClipboard = "@@Eisbuk/DELETE_SLOT_FROM_CLIPBOARD0",
   AddSlotToClipboard = "@@Eisbuk/ADD_SLOT_TO_CLIPBOARD",
 
-  IsOrganizationStatusReceived = "@@Eisbuk/IS_ORGANIZATION_STATUS_RECEIVED",
+  IsOrganizationStatusReceived = "@@Eisbuk/IS_getOrganization()_STATUS_RECEIVED",
```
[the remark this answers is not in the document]
fadwamahmoud: Also a mistake. Find and replace was involved...

ikusteu (approved):

PR #242, opened 2021-11-17
Cypress tests ci

silviot:


eisbuk-admin/src/store/actions/firestoreOperations/subscriptionHandlers.ts
```diff
@@ -198,7 +200,7 @@ export const subToAttendance: SubscriptionHandler = ({
 }) => {
   const collRef = collection(
     getFirestore(),
-    `${orgPath}/${OrgSubCollection.Attendance}`
+    `${getOrgPath()}/${OrgSubCollection.Attendance}`
```
ikusteu, line 203: To make it prettier, you can do `collection(getFirestore, getOrgPath(), ...<theRestOfThePathElements>)` since it utilizes the `...rest` operator (sort of like a `path.join()`)

ikusteu (approved):

eisbuk-admin/src/store/actions/firestoreOperations/subscriptionHandlers.ts
silviot: Thanks for the suggestion! Done.

PR #243, opened 2021-11-18
Cypress tests

fadwamahmoud:
E2e Cypress tests for the following:
- logging in flow.
- creating an athlete.
- creating a slot with multiple scenarios.
- some additional scenarios are yet to be included in creating an athlete.


silviot:
I moved the tests from here into #250 .
I'm closing this now. @fadwamahmoud Feel free to point out things present here that still need to be included in the project, in case I missed something.

PR #244, opened 2021-11-18
Update/firebase@9.1.1

### LINKED ISSUE #133: Update firebase to v9
Version 9 of the firebase client library has a new structure that enables tree shaking.
See https://firebase.google.com/docs/web/learn-more#modular-version

Should probably happen after #115 to solve not too many issues at the same time.

- [x] Create `react-redux-firebase`-like listener subscriptions
- [x] Rewrite SDK syntax from firebase v8 to v9
- [x] Create auth logic
- [x] Create login flow (replace `react-firebase-ui`)

### LINKED ISSUE #237: Add admin and non-admin firestore instances for tests
Currently we're bypassed the testing of admin, non admin. We should revisit that and create two instances for testing: 
- regular db
- admin db

- [x] Create two (firestore, or just auth) instances for testing
- [x] Apply to already existing (but skipped/commented out) auth tests

ikusteu:
Fixes #237 
Fixes #133 

This should not be merged yet, however, it's complete, so please do review it/hand test it. I would like to now go over the other PRs, merge and rebase this one with all of the updates.

- The firebase is updates to v9 with all of the API changed to use the treeshakable (v9) syntax
- The `react-redux-firebase` library has been replaced with our own logic (resembling `react-redux-firebase`) because the library doesn't currently support v9 syntax and breaks on imports due to breaking changes between v8 and v9
- The tests are updated accordingly so that `adminDb` is used ONLY when needed, and all the rest is tested with client SDKs firestore instances


eisbuk-admin/.babelrc.js
```diff
@@ -1,4 +1,5 @@
 const plugins = [
+  ["@babel/plugin-proposal-private-methods", { loose: true }],
```
silviot, line 2: Thanks! These warnings were about to drive me crazy!

ikusteu: :wink: did it just now :sunglasses:

silviot (approved):

PR #247, opened 2021-11-21
Feature/husky pre commit

ikusteu:
I've replaced (python based) `pre-commit` with husky pre-commit + lint-staged hooks. The `pre-commit` job in `build` workflow was entirely replaced with `run-lint` job for more accurate/compatible checks.

Reasons for this transition:
1. The linting is easier to set up as we can use linting/prettier scripts set up in `lint-staged`, utilizing appropriate `.eslintrc.js` configs instead of writing eslint deps and rules in `.pre-commit-config.yaml` (easier use, less redundancy)
2. The linting was broken at times and this is easier to debug (due to benefits from 1. point)
3. This way we're only using shell scripts and node processes, without python, reducing one additional dependency for everyone's local environment (python)

Some tradeoffs, todos:
0. ~~We're not checking yaml files~~
1. There is no end of line/file fixing present: the EOF and EOL are fixed with prettier command with `lint-staged` scripts, but it only works on JSON, js/ts files, not on each file (such as .yaml .rules .git, etc)
_I might write my own scripts for that (in Go), which will be compiled to CLI executables, not needing Go to run, or python (like pre-commit does). I would do it on my own time, but only if necessary_


.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
```
silviot, line 2: How should this script be used?

silviot:
`husky` is installed in `eisbuk-admin/package.json` and there's currently no documentation about how to set up the git hook locally. I propose we create a new `package.json` at the top level including `husky`, and use that for the git hook.

silviot:
I went on with the proposed changes and created a top level `package.json`.
The `eisbuk-install.sh` script is no longer needed. Now [yarn workspaces](https://yarnpkg.com/features/workspaces) take care of installing dependencies for all our packages.

silviot (approved):

PR #249, opened 2021-11-24
Create 'initAdminApp' cypress Command and fix initial tests

ikusteu:
I've created an `initAdminApp` cypress Command. It does all of the functionality we want before each test:
- creates a new (random) organization string and stores it in local storage
- creates said organization in firestore
- creates/signs in default user (already added as admin in default organization)

Other than that, I've applied some small fixes where the tests were faulty and replaced all standardized strings (routes, category, slot type) with appropriate enums.


eisbuk-admin/cypress/integration/create_slot_spec.ts
```diff
@@ -53,7 +22,7 @@ describe("Create slot", () => {
         .should("have.attr", "aria-disabled", "true")
         .and("have.attr", "aria-current", "page");
 
-      cy.get("a[href='/']")
+      cy.get(`a[href='${PrivateRoutes.Slots}']`)
```
silviot, line 25: This is nice!

silviot (approved):

PR #250, opened 2021-11-25
Add login test

silviot:
Add the login test from #243


eisbuk-admin/cypress/integration/create_slot_spec.ts
```diff
@@ -92,4 +87,35 @@ describe("Create slot", () => {
     cy.get("[aria-label='Additional slot notes']").type("some notes");
     cy.get("[aria-label='Confirm slot creation']").click();
   });
+
+  it("submits an invalid time format", () => {
```
ikusteu, line 91: I would write this as "showa 'invalid time format' invalidation error" (doesn't technically submit)

eisbuk-admin/cypress/integration/create_slot_spec.ts
```diff
@@ -92,4 +87,35 @@ describe("Create slot", () => {
...
+    cy.visit(PrivateRoutes.Slots);
+    cy.get("[type='checkbox']").click();
+    cy.get("[data-testid='new-slot-button']").first().click();
+    cy.get("[value='ice']").check();
+    cy.get("[value='competitive']").check();
```
ikusteu, line 96: I would replace these with enums (`SlotType` and `Category`)

eisbuk-admin/cypress/integration/create_slot_spec.ts
```diff
@@ -92,4 +87,35 @@ describe("Create slot", () => {
...
+
+    cy.get("[type='submit']").click();
+    cy.contains("Invalid time format");
+  });
+  it("submits an invalid interval", () => {
+    cy.visit(PrivateRoutes.Slots);
+    cy.get("[type='checkbox']").click();
+    cy.get("[data-testid='new-slot-button']").first().click();
+    cy.get("[value='ice']").check();
+    cy.get("[value='competitive']").check();
```
ikusteu, line 109: Enums

eisbuk-admin/cypress/integration/create_slot_spec.ts
```diff
@@ -92,4 +87,35 @@ describe("Create slot", () => {
...
+
+    cy.get("[type='text']").eq(0).clear().type("9");
+    cy.get("[type='text']").eq(1).clear().type("10 pm");
+
+    cy.get("[type='submit']").click();
+    cy.contains("Invalid time format");
+  });
+  it("submits an invalid interval", () => {
```
ikusteu, line 104: "shows 'time mismatch' (or whatever) validation error"

eisbuk-admin/cypress/support/commands.js
```diff
@@ -36,28 +37,41 @@ connectFunctionsEmulator(functions, "localhost", 5001);
...
  * Set up app for testing and log in as default admin
+ * @param {boolean} doLogin - whether to log in as default admin
  */
-Cypress.Commands.add("initAdminApp", () => {
+Cypress.Commands.add("initAdminApp", async (doLogin = true) => {
   // create a random organization name in order to run each test
   // against it's own organization, using the same db without conflicts
   const organization = uuidv4();
+  console.log(doLogin);
+  debugger;
```
ikusteu, line 47: I would either remove the `console.log` (if it was a WIP debug check), or make it more verbose if it's part of standard reporting. I don't understand what the debugger does here, to me it seems as just an uncalled expression.

eisbuk-admin/cypress/support/commands.js
```diff
@@ -36,28 +37,41 @@ connectFunctionsEmulator(functions, "localhost", 5001);
...
+        auth,
+        defaultUser.email,
+        defaultUser.password
+      );
+      console.log(`Logged in as ${defaultUser.email}`);
+    } else {
+      debugger;
+      await signOut(auth);
+      console.log(`Logged out`);
+    }
```
ikusteu, line 75: I don't quite understand how this works...From my understanding, the login should be ran only if the `createUserWithEmailAndPassword` throws (the user already exists). Following that logic, if createUser(...) is successful, the `if (doLogin)` won't be ran and (from my understanding) the created user will remain logged in (as no `signOut` is ran)...but then again the (UI) login test would not be possible after this, but still it passes...

Could you please explain this

ikusteu (changes requested):
I left some nitpicky semantics change requests.

eisbuk-admin/cypress/support/commands.js
silviot: This is what code reviews are for! I forgot to take these out.

eisbuk-admin/cypress/support/commands.js
silviot: Right! It was working because the non-catch path only happens on the first test, since subsequent tests will find an already existing user.
I fixed the function converting the `.catch()` to use `try {} catch()` instead. The problem you mentioned should be fixed, and it's now more readable. There was also a spurious `debugger` statement.

eisbuk-admin/cypress/support/commands.js
```diff
@@ -36,28 +37,43 @@ connectFunctionsEmulator(functions, "localhost", 5001);
...
+      await signInWithEmailAndPassword(
+        auth,
+        defaultUser.email,
+        defaultUser.password
+      );
+      console.log(`Logged in as ${defaultUser.email}`);
+    } else {
+      await signOut(auth);
+      console.log(`Logged out`);
+    }
```
ikusteu, line 77: This seems unnecessary as the user will not be signed in if the create user failed. We should only run login if needed (to log in a user who, obviously, exists)

silviot: That was my thought as well, but I was not able to make it work that way.
I wonder if cypress is not cleaning everything up between tests.
Or maybe I saw something that was not there.
If it works without this, feel free to take it out.

eisbuk-admin/cypress/integration/create_slot_spec.ts
```diff
@@ -58,38 +56,75 @@ describe("Create slot", () => {
...
+    cy.getAttrWith("data-testid", "new-slot-button").first().click();
+    cy.getAttrWith("value", SlotType.Ice).check();
+    cy.getAttrWith("value", Category.Competitive).check();
+
+    cy.getAttrWith("type", "text").eq(0).clear().type("9:00");
+    cy.getAttrWith("type", "text").eq(1).clear().type("7:00");
+
+    cy.getAttrWith("type", "submit").click();
+    // The dialog is still there
+    cy.getAttrWith("role", "dialog").within(() => {
```
silviot, line 125: Can we chain the `contains` to the `getAttrWith` without using `within`?

eisbuk-admin/cypress/support/commands.js
delasous: Have left this.

.github/workflows/build.yml
```diff
@@ -166,13 +169,14 @@ jobs:
       - name: Cypress run
         uses: cypress-io/github-action@v2
         with:
-          start: yarn emulators:start
+          start: yarn emulators:start; yarn cypress:logger
```
silviot, line 172: I think the separator should be a comma here. I believe the cypress logger might not run at all if a `;` is used.

eisbuk-admin/src/components/atoms/SlotOperationButtons/CopyButton.tsx
```diff
@@ -76,7 +76,7 @@ export const CopyButton: React.FC<SlotButtonProps> = ({ size }) => {
         size={size || iconSize}
         onClick={onCopy}
         data-testid={__copyButtonId__}
-        aria-label="Copy slots button"
```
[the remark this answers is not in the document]
delasous: Ok, noted going forward. This has been resolved in translations commits below

fadwamahmoud (approved):

ikusteu (approved):

PR #252, opened 2021-11-29
Fix production deploy

silviot:
Production was broken. It had the `__isDev__` flag set to `true`.
The reason for this was the condition inspecting the `BUILD_ENV`
variable did a strict comparison against the empty string:
it should be a dev build if and only if the `BUILD_ENV` variable
is different from the empty string.

But since we don't provide a value to the `BUILD_ENV` variable in the CI
job that does the deployment, `process.env.BUILD_ENV` ended up
having the value `undefined`, and a dev build was being pushed to production.

This PR changes the condition to check if the variable has a truthy value.
This should be what we want in this case: a non empty string will yield a
`true` while an `undefined` or an empty string will yield a `false`.
See https://developer.mozilla.org/en-US/docs/Glossary/Truthy

I also sneaked in a simplification: we can pass environment variables to the build step instead of stuffing them in a `.env` file, saving a couple of steps in the job.

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -24,21 +24,16 @@ jobs:
           key: ${{ runner.os }}-modules-v12-${{ hashFiles('**/yarn.lock') }}
       - name: Install and link packages
         run: yarn
-      - name: Create .env file with credentials
-        uses: SpicyPizza/create-envfile@v1
-        with:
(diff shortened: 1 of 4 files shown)


silviot:
Cypress tests are still broken, and I don't understand exactly why.
@ikusteu Can you take a look when you have a minute?

Note that since deployment to production happens on merges to `master` and is currently broken, this PR should go in before any other does.

silviot:
I merged these changes to #250

PR #254, opened 2021-12-01
Fix/athletes grid

### LINKED ISSUE #233: The athlets grid should be as wide and as tall as the page
We probably have one component too many in the hierarchy.
The arrows point at unnecessary whitespace:
image

fadwamahmoud:
Fixes #233 
- Removed container wrapper from `customersList` component which was hindering the default athletes grid from expanding.
- Created new `classNames` prop to accept multiple classes.
- Provided different `tableContainer` classes according to where it's called.

### DIFF
--- a/eisbuk-admin/src/components/atoms/CustomerList/CustomerList.tsx
+++ b/eisbuk-admin/src/components/atoms/CustomerList/CustomerList.tsx
@@ -23,14 +23,14 @@ interface Props {
   customers?: Customer[];
   extended?: boolean;
   onCustomerClick?: (customer: Customer) => void;
-  className?: string;
+  classNames?: { rootDiv?: string; tableContainer?: string };
 }
 
 const CustomerList: React.FC<Props> = ({
   customers,
   extended,
   onCustomerClick = () => {},
-  className = "",
+  classNames = { rootDiv: "" },
 }) => {
   const classes = useStyles();
   const { t } = useTranslation();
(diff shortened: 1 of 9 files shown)


silviot (approved):
LGTM

eisbuk-admin/src/components/atoms/CustomerList/CustomerList.tsx
```diff
@@ -23,14 +23,14 @@ interface Props {
   customers?: Customer[];
   extended?: boolean;
   onCustomerClick?: (customer: Customer) => void;
-  className?: string;
+  classNames?: { rootDiv?: string; tableContainer?: string };
```
ikusteu, line 26: Right...this is generally a bad practice to go about it this way:
- you shouldn't let the outside component (explicitly) control too much of the internals of a component 
- the convention of passing the `className` as a string should always be respected, and `className` should (in most cases) be applied to the root container of the component (in this case the outermost div element)
- in edgy edge case, you might want to do it and you should consider two things:
1. Leave the `className` as is (string applied to the root element) and additional `tableContainerClassName` (MaterialUI does this for some components)
2. As @silviot says it: put a 5$ in a jar each time you do it

In this case, I would actually go with the non-optimal option and pass them temporarily as `className` and `tableContainerClassName`, but add `/** @TODO fix scrolling */` because this is still a temporary solution (which I've added for the time being) that doesn't work for all cases (small screens) adn should be revisited.

eisbuk-admin/src/components/atoms/CustomerList/CustomerList.tsx
```diff
@@ -48,9 +48,11 @@ const CustomerList: React.FC<Props> = ({
   const searchRegex = new RegExp(searchString, "i");
 
   return (
-    <div className={className}>
+    <div className={classNames?.rootDiv}>
```
ikusteu, line 51: Generally, this `?` is unnecessary as you have a fallback, so `classNames` will always be defined. The problem could arise if someone passes `classNames={ }`, overriding your fallback, but `classNames.rootDiv` will be `undefined` (and it should be `string`). In that case, this will still not fix it and is one of the reasons for my comment above: when you pass two (optional) strings, instead of an object, it's easier (and more straightforward) to provide a fallback, catching all of the cases.

Issue #169, opened 2021-08-26
Birthday notification


silviot:
There should be something in the app that tells admins if today is any athlete's birthday.
It could be an icon that is grey when it's nobody's birthday, and is coloured when it is. In the latter case when clicked it should show a list of athletes whose birthday is today.

PR #257, opened 2021-12-08
Feature/birthday notifications

fadwamahmoud:
Fixes: #169 
- BirthdayMenu component is created in AppbarAdmin to display athletes sorted by their birthday
- A new selector is created to fetch customers from store and sort them by their birthday plus tests
- Some improvements are yet to be made 
         1. Using the CustomersList component  
         2. Writing tests for BirthdayMenu 
         3. Using timeago.js to display today and tomorrow instead of hardcoding them

### DIFF
--- a/eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.stories.tsx
+++ b/eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.stories.tsx
@@ -8,7 +8,7 @@ export default {
   title: "Birthday Menu",
   component: BirthdayMenu,
 };
-const customers = { "12-12-2020": [saul, walt] };
+const customers = [{ birthday: "2020-12-12", customers: [saul, walt] }];
 export const Default = (): JSX.Element => (
   <BirthdayMenu customers={customers} />
 );
--- a/eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.stories.tsx
+++ b/eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.stories.tsx
@@ -1,9 +1,10 @@
 import React from "react";
 
+import { CustomersByBirthday } from "eisbuk-shared";
+
 import BirthdayMenu from "./BirthdayMenu";
 
 import { saul, walt } from "@/__testData__/customers";
-import { CustomersByBirthday } from "eisbuk-shared/dist";
 
 export default {
   title: "Birthday Menu",
--- a/eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.stories.tsx
+++ b/eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.stories.tsx
@@ -8,7 +8,7 @@ export default {
   title: "Birthday Menu",
   component: BirthdayMenu,
 };
-
+const customers = { "12-12-2020": [saul, walt] };
 export const Default = (): JSX.Element => (
-  <BirthdayMenu customers={[saul, walt]} />
+  <BirthdayMenu customers={customers} />
 );
--- a/eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.stories.tsx
+++ b/eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.stories.tsx
@@ -3,12 +3,15 @@ import React from "react";
 import BirthdayMenu from "./BirthdayMenu";
 
 import { saul, walt } from "@/__testData__/customers";
+import { CustomerBirthday } from "eisbuk-shared/dist";
 
 export default {
   title: "Birthday Menu",
   component: BirthdayMenu,
 };
-const customers = [{ birthday: "2020-12-12", customers: [saul, walt] }];
+const customers: CustomerBirthday[] = [
+  { birthday: "2020-12-12", customers: [saul, walt] },
+];
 export const Default = (): JSX.Element => (
   <BirthdayMenu customers={customers} />
 );
--- a/eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.stories.tsx
+++ b/eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.stories.tsx
@@ -1,0 +1,14 @@
+import React from "react";
+
+import BirthdayMenu from "./BirthdayMenu";
+
+import { saul, walt } from "@/__testData__/customers";
+
+export default {
+  title: "Birthday Menu",
+  component: BirthdayMenu,
+};
+
+export const Default = (): JSX.Element => (
+  <BirthdayMenu customers={[saul, walt]} />
+);
--- a/eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.stories.tsx
+++ b/eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.stories.tsx
@@ -3,13 +3,13 @@ import React from "react";
 import BirthdayMenu from "./BirthdayMenu";
 
 import { saul, walt } from "@/__testData__/customers";
-import { CustomerBirthday } from "eisbuk-shared/dist";
+import { CustomersByBirthday } from "eisbuk-shared/dist";
 
 export default {
   title: "Birthday Menu",
   component: BirthdayMenu,
 };
-const customers: CustomerBirthday[] = [
+const customers: CustomersByBirthday[] = [
   { birthday: "2020-12-12", customers: [saul, walt] },
 ];
 export const Default = (): JSX.Element => (
--- a/eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.tsx
+++ b/eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.tsx
@@ -51,7 +51,7 @@ const BirthdayMenu: React.FC<Props> = ({ customers }) => {
               <MenuItem>
                 {customerBirthday === today ? "today" : customerBirthday}
               </MenuItem>
-              <Divider></Divider>
+              <Divider />
               {customer.customers.map(
                 (cus) =>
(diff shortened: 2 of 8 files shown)


eisbuk-admin/src/store/selectors/customers.ts
```diff
@@ -22,3 +23,41 @@ export const getCustomersList = (state: LocalStore): Customer[] => {
...
+    (c) =>
+      DateTime.now().startOf("day").set({ year: 2000 }) <=
+      DateTime.fromISO(c.birthday).startOf("day").set({ year: 2000 })
+  );
+  const rearrangedCustomer = sortedCustomers
+    .slice(index === -1 ? 0 : index)
+    .concat(sortedCustomers.splice(0, index));
+
+  return rearrangedCustomer.reduce(
+    (r, v, i, a, k = DateTime.fromISO(v.birthday).toFormat("dd/MM")) => (
```
silviot, line 57: ISO dates are `YYYY-mm-dd`. We don't need to parse them and format them to get rid of the first five charaters. `v.birthday.substring(5)` should be enough.
Also I'm not a fan of single letter variables, unless they are `i` and `j`, and they're integers representing the index of a loop, or [a few other cases](https://www.codereadability.com/i-n-k-single-letter-variable-names/).

eisbuk-admin/src/store/selectors/customers.ts
```diff
@@ -22,3 +23,41 @@ export const getCustomersList = (state: LocalStore): Customer[] => {
...
+    DateTime.fromISO(a.birthday)
+      .set({ year: 2000 })
+      .toString()
+      .localeCompare(
+        DateTime.fromISO(b.birthday).set({ year: 2000 }).toString()
+      )
+  );
+  const index = sortedCustomers.findIndex(
+    (c) =>
+      DateTime.now().startOf("day").set({ year: 2000 }) <=
```
silviot, line 49: This function ideally should be pure, i.e. its result should only depend on the passed parameters. Striving to have as many pure functions as possible results in an easier to maintain codebase, since side effects are contained.

For this reason it would be better not to invoke `now` here, but rather rely on a date passed by the caller. The UI code that will use this function will be responsible of invoking `now()` and passing the information down here.

silviot (dismissed):

eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.stories.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+import BirthdayMenu from "./BirthdayMenu";
+
+import { saul, walt } from "@/__testData__/customers";
+
+export default {
+  title: "Birthday Menu",
+  component: BirthdayMenu,
+};
+const customers = { "12-12-2020": [saul, walt] };
```
ikusteu, line 11: This should be an ISO string (as it will be so in production)

eisbuk-admin/src/store/selectors/__testData__/customers.ts
```diff
@@ -0,0 +1,24 @@
...
+import { createTestStore } from "@/__testUtils__/firestore";
+import { saul, walt, jian } from "@/__testData__/customers";
+
+/**
+ * A customers store input
+ */
+const customers: LocalStore["firestore"]["data"]["customers"] = {
+  [saul.secretKey]: saul,
+  [jian.secretKey]: jian,
+  [walt.secretKey]: walt,
```
ikusteu, line 12: This will not change the behavior of tests/display, but customers are keyed by id, not secret key, so it would make sense to keep things consistent here as well.

eisbuk-admin/src/store/selectors/customers.ts
ikusteu: I agree here (even though I use them extensively in Go), and they will be compiled down to single letter vars on bundling (to reduce code as much as possible), but that reduction (bundle) isn't meant to be read by human, but rather just the JS interpreter.

eisbuk-admin/src/store/selectors/customers.ts
```diff
@@ -22,3 +23,41 @@ export const getCustomersList = (state: LocalStore): Customer[] => {
...
+      DateTime.fromISO(c.birthday).startOf("day").set({ year: 2000 })
+  );
+  const rearrangedCustomer = sortedCustomers
+    .slice(index === -1 ? 0 : index)
+    .concat(sortedCustomers.splice(0, index));
```
ikusteu, line 54: Each time you use splice you need to put 5$ in a jar: slice will return the subslice (array) of the passed array, while splice will mutate the passed array (like push/pop).

eisbuk-admin/src/store/selectors/__testData__/customers.ts
```diff
@@ -0,0 +1,24 @@
...
+};
+
+export const expectedCustomersBirthdays = {
+  "01/01": [saul, jian, walt],
```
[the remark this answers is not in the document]
ikusteu: I wouldn't use the `mm/dd` format (or vice versa), but rather the standard ISO `mm-dd` as keys. The conversion for displaying should be done at component level: if you want to display `31.10` instead of `10-31`, but still use `10-31` to process/key/query the data.

eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.tsx
```diff
@@ -0,0 +1,72 @@
...
+        }}
+      >
+        {customers.map((customer) => {
+          const customerBirthday = customer.birthday.substring(5);
+          return (
+            <div key={customer.birthday}>
+              <MenuItem>
+                {customerBirthday === today ? "today" : customerBirthday}
+              </MenuItem>
+              <Divider></Divider>
```
ikusteu, line 54: This can just be `<Divider />`

eisbuk-admin/src/components/layout/AppbarAdmin.tsx
```diff
@@ -134,6 +140,8 @@ const AppbarAdmin: React.FC<AppBarProps> = (props) => {
               <MenuIcon />
             </Button>
           </Hidden>
+
+          <BirthdayMenu customers={customers} />
```
ikusteu, line 144: Why the excess spaces?

eisbuk-admin/src/store/selectors/customers.ts
```diff
@@ -22,3 +23,46 @@ export const getCustomersList = (state: LocalStore): Customer[] => {
   const customersInStore = getCustomersRecord(state);
   return Object.values(customersInStore);
 };
+
+/**
+ * Get a list of all the customers whose birthdays are today
+ * (not an actual firestore query but rather synced local store entry)
+ * @param state Local Redux Store
+ * @returns list of customers grouped by birthday
```
ikusteu, line 31: This actually creates a selector, so the only `@param` should be `date`

eisbuk-admin/src/store/selectors/customers.ts
```diff
@@ -22,3 +23,46 @@ export const getCustomersList = (state: LocalStore): Customer[] => {
...
+        .toString()
+        .localeCompare(
+          DateTime.fromISO(b.birthday).set({ year: 2000 }).toString()
+        )
+    );
+    const index = sortedCustomers.findIndex(
+      (c) =>
+        date.startOf("day").set({ year: 2000 }) <=
+        DateTime.fromISO(c.birthday).startOf("day").set({ year: 2000 })
+    );
```
ikusteu, line 62: This could just be string comparison (substr(5)) which will give you `mm-dd` from `yyyy-mm-dd` and then two of those are comparable (as dates within a year)

eisbuk-admin/src/store/selectors/customers.ts
```diff
@@ -22,3 +23,46 @@ export const getCustomersList = (state: LocalStore): Customer[] => {
...
+        .set({ year: 2000 })
+        .toString()
+        .localeCompare(
+          DateTime.fromISO(b.birthday).set({ year: 2000 }).toString()
+        )
+    );
+    const index = sortedCustomers.findIndex(
+      (c) =>
+        date.startOf("day").toString().substring(5) <=
+        DateTime.fromISO(c.birthday).startOf("day").toString().substring(5)
```
ikusteu, line 61: Right, so...no `startOf` setting... you get a birthday string, for instance `"2021-08-01"` and a different ISO string to compare, for instance "2020-05-22". Then you just compare the two: 
```typescript
const compareDateOfYear = (a: string, b: string) => a.substr(5) <= b.substr(5) 
```
You can use a version of this for each date comparison within the selector, so I suggest you either convert the `date: DateTime` input to ISO string at the beginning of the selector and just use the `dateISO` or require the `date` param to be `date: string` (and do the conversion outside the selector)

eisbuk-admin/src/store/selectors/customers.ts
```diff
@@ -22,3 +23,44 @@ export const getCustomersList = (state: LocalStore): Customer[] => {
...
+          });
+    });
+    const sortedCustomers = customers.sort((a, b) =>
+      DateTime.fromISO(a.birthday)
+        .set({ year: 2000 })
+        .toString()
+        .localeCompare(
+          DateTime.fromISO(b.birthday).set({ year: 2000 }).toString()
+        )
+    );
```
ikusteu, line 57: This should be done with string comparison as well

ikusteu (approved):

eisbuk-admin/src/store/selectors/__testData__/customers.ts
```diff
@@ -0,0 +1,38 @@
...
+  [saul.id]: saul,
+  [jian.id]: jian,
+  [walt.id]: walt,
+  [mike.id]: mike,
+  [jane.id]: jane,
+};
+
+export const expectedCustomersBirthdays = [
+  {
+    birthday: "2009-12-23",
```
ikusteu, line 20: We're expecting "12-23" to be the birthday as customers born on "2021-12-23" and "1994-12-23" will have the same birthday

PR #262, opened 2021-12-12
Feature/esbuild client

### LINKED ISSUE #199: Annihilate react-hooks/exhaustive-deps
It is a rule which we've found to bring more cost than benefit and would like to ignore it. However, as much as `.eslintrc.js` is updated, the CI eslint always throws that error (on pre-commit), we have a temporary solution to ignore it for  file/line, but this should be done clean:
- [x] Apply ignore for said rule in `.eslintrc.js` and check that it works with CI
- [x] Remove all ignore-for-line comments

_note: this might not be an eslint problem, but pre-commit and might be solved with #129_

### LINKED ISSUE #258: Replace Webpack with ESBuild for client app
After migrating `/functions` bundling from webpack to esbuild, it makes sense to do the same for the rest of the codebase. The next logical step would be to bundle the main app using esbuild and provide the dev server based on esbuild as well (due to lightning fast build times):

- [x] Use ESBuild to bundle the app (replace `build` script)
- [x] Create `build:test` script to build the app with `NODE_ENV ===  "test"`
- [x] Create ESBuild based dev server (replace `serve` script)
- [x] Create a logic for api fallback (mainly for dev server) -> implement a catch all serve proxy
- [x] Add refresh signal to dev server after the changes have been noticed (and app rebuilt)
- [x] Add type check in CI (since it's not done at build time)

ikusteu:
Fixes #258 
Fixes #199 
The build times are now super fast (1.4-1.6s), dev server is ok..the refresh signal has been implemented using esbuild's `watch.onRebuild` and SSE message. The issue with cypress has been solved (inconsistent api key) and now all of the checks are passing.


eisbuk-admin/esbuild/src/lib/config.ts
```diff
@@ -0,0 +1,19 @@
+import { BuildOptions } from "esbuild";
+import path from "path";
+import svgr from "esbuild-plugin-svgr";
+
+/**
+ * A base esbuild config file shared between both `build` and `serve`
+ * scripts.
+ */
+const config: BuildOptions = {
+  entryPoints: [path.join(process.cwd(), "src", "index.tsx")],
```
silviot, line 10: Using a path relative to the current file would work even when the build is invoked from an unusual directory.

silviot: Nevermind:

* the current directory is changed by yarn when invoking `yarn build` to be `eisbuk-admin`
* the `config.ts` file is executed by reading it into a buffer and running `eval` on it, so we don't really have access to `__dirname` or `__filename` there

eisbuk-admin/package.json
```diff
@@ -87,7 +89,7 @@
     "@types/jest": "^26.0.23",
     "@types/lodash": "^4.14.170",
     "@types/luxon": "^2.0.5",
-    "@types/node": "^12.0.0",
+    "@types/node": "^16.11.10",
```
silviot, line 92: We should upgrade everything to node 16. Currently our `.nvmrc` states `v14.18.1` and `functions/package.json` specifies `14`. Node 16 in Firebase functions is supported as a beta release; we should upgrade to that version.

Not so much a comment about this work, just mentioning.

eisbuk-admin/esbuild/src/serve.ts
```diff
@@ -0,0 +1,106 @@
...
+    const open = {
+      darwin: ["open"],
+      linux: ["xdg-open"],
+      win32: ["cmd", "/c", "start"],
+    } as Record<NodeJS.Platform, string[]>;
+    const ptf = process.platform;
+    if (clients.length === 0) {
+      spawn(open[ptf][0], [...open[ptf].slice(1), `http://localhost:3000`]);
+    }
+  }, 1000);
```
silviot, line 99: The arbitrary value "one second" will be more than needed most of the time (no big deal - we lose a few milliseconds of our lives waiting for a browser to open) and less than needed from time to time (this is more of a problem).
The thing we should wait for is a `200` code when requesting `/app/bundle.js`: when that happens we're ready to open the browser. I'll push this change.

silviot (approved):
Very nice! It's refreshing to use!

PR #263, opened 2021-12-13
Fix scrolling for customer list

### LINKED ISSUE #255: Fix up CustomerList scrolling (overflow)
We need to allow scrolling in `CustomerList` component, for smaller screens. This is used in production within `AddCustomers` component (pop-up) of `AttendanceCard`. Currently the scrolling is controlled by passing additional className to one of the internal elements of component (`tableContainerClassName`), which is not ideal. 
Additionally, the scrolling utilizes `maxHeight` which doesn't work in all cases (on small screens) as the `maxHeight` overflows the container component, in other works the entire scroll container overflows it's container so part of the scroll container is invisible (and it's contents unreachable).

To fix this we should first try to make it work with MUI, but it might make our lives a bit easier to create a HTML+CSS component, which won't be that complex, but would give us a complete control over the component (as opposed to MUI)

- [x] Fix the scrolling so that it works on all screens (avoid `maxHeight`)
- ~Make the classes on table container internal -> only control from outside should be the `variant`~

ikusteu:
Fixes #255 
I've fixed the scrolling issue in the CustomerList component (both add customer and `/customers` display). The fix is somewhat suboptimal: the component doesn't control it internally, but we rather need to set explicit `maxWidth` through `tableContainerClassName`, but I'm willing to put $5 in that jar as it seems as the best option we have in this case.

### DIFF
--- a/eisbuk-admin/src/components/atoms/AttendanceCard/AddCustomers.tsx
+++ b/eisbuk-admin/src/components/atoms/AttendanceCard/AddCustomers.tsx
@@ -115,8 +115,9 @@ const useStyles = makeStyles((theme) => ({
     backgroundColor: "rgba(255, 255, 255, 0.6)",
   },
   tableContainer: {
-    overflow: "auto",
-    maxHeight: "30rem",
+    overflowY: "auto",
+    overflowX: "hidden",
+    maxHeight: "calc(100vh - 14rem)",
   },
 }));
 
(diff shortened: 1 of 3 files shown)


fadwamahmoud (approved):

silviot:
On https://igorice--263-merge-di14hzs9.web.app/athletes I can't horizontally scroll anymore.
It's probably ok though, since requiring the user to do horizontal scrolling is not a good UX pattern, and we have plans to improve the athletes view anyway.

silviot (approved):

PR #268, opened 2021-12-21
Feature/secret key email

### LINKED ISSUE #101: Button to send athletes a link to manage their bookings : 13 pts
Under athletes listing show a button for each athlete that will send them a text-only email that includes a link to manage their bookings.

We need a function that takes an athlete id, looks up their email and performs the sending.

Code for sending can use https://nodemailer.com/about/ like the extension  https://github.com/firebase/extensions/tree/master/firestore-send-email

Use values taken from #102 for the SMTP server. In case they're absent use `localhost:2500` (useful while testing).
Debugging of this feature can then be aided by mailslurper running  
```
docker run -it -p 2500:2500 -p 8080:8080 -p 8085:8085 --rm marcopas/docker-mailslurper@sha256:6ffa582222e0c52c0db325ba2b67e67c097f37ddf901f31540f5e19376d8a165
```

and going to http://localhost:8080/

ikusteu:
Fixes #101 

The email sending logic is currently implemented with the help of Firebase's `firestore-send-email` extension. The extension works as a data trigger (further handling the email logic) each time we write to `mail` collection. The in-app functionality handles the writes to the DB, whereas the extension is in charge of email sending logic and without said extension installed/activated, the `sendEmail` (in-app) cloud function would merely write to the `mail` collection, without the email being sent

To fully set up the extension, it should first be installed in firebase console, there the extension should be configured to the appropriate SMTP server with the right credentials. The default collection (`mail`) is compatible with the app by default but can be changed: both in extension config, and by simply changing the value of enum `Collection.EmailQueue` to the desired name.


silviot:
A comment after a first view: the email will need some tweaks (from address, a better template) and this can be done in a later step.
But there's one thing that concerns me: you might accidentally click on a "send email" button and trigger an unwanted email. I think adding a confirmation dialog is the easiest way to prevent this.

eisbuk-admin/firestore.rules
```diff
@@ -37,16 +37,17 @@ service cloud.firestore {
         );
     }
     match /organizations/{organization} {
+      allow create, update: if isAdmin(organization);
```
silviot, line 40: Note to self: administrators have permission to remove themselves from an organization. Not a problem, just keeping it in mind. We don't even have any UI yet to manage admins.

eisbuk-admin/functions/src/utils.ts
```diff
@@ -71,7 +78,7 @@ const isOrgAdmin = (admins: string[] | undefined, auth: Auth): boolean => {
 const throwUnauth = (): never => {
   throw new functions.https.HttpsError(
     "permission-denied",
-    "unauthorized",
+    HTTPErrors.Unauth,
```
silviot, line 81: Nice!

eisbuk-admin/package.json
```diff
@@ -21,6 +21,7 @@
     "i18next-http-backend": "^1.2.6",
     "luxon": "^2.0.2",
     "moment": "^2.27.0",
+    "node-mailer": "^0.1.1",
```
silviot, line 24: We probably don't need this if the trigger will take care of email sending.

eisbuk-admin/src/store/actions/customerOperations.ts
```diff
@@ -79,28 +83,60 @@ export const updateCustomer = (
...
+  accessLink: string;
+}
+
+export const sendBookingsLink =
+  ({ to, accessLink, subject }: BookingMailProps): FirestoreThunk =>
+  async (dispatch) => {
+    try {
+      const newEmail: EmailMessage = {
+        to,
+        message: { subject, html: accessLink },
```
silviot, line 124: We'll need to add some template here to put some boilerplate in the email. It can go in as is for now, but this is something we'll need soon.

The template needs access to the athlete object to be able to write `Dear John Doe, here is your link`. So either we pass the full message to this function (in this case the `accessLink` should be renamed to `emailBody`) or we pass the athlete object too so that this function has all it needs to compose the email.

ikusteu: I'm more down with the second solution. Instead of passing `accessLink`, we will pass customer id and the rest can be generated from store entry for customer keyed by customer id.

On it!

silviot:
Merging #272 made this PR not clean. You'll need to either rebase (my personal favourite) or merge the `master` branch.

silviot (approved):

Issue #266, opened 2021-12-20
Birthday notification follow up


ikusteu:
Birthday notification component looks nice, but there is some stuff that would be nice to have (extend the functionality)
- [ ] Create a customer birthdays page (on which we can see all of the customer birthdays)
- [x] Show only limited number of customers in a popup and provide a "see all" button redirecting to customer birthdays page
- [x] Make the dates (of birthdays) a bit prettier (applying i18n format should do)
- [ ] Show a badge (like facebook notification) with number of birthdays this week

PR #272, opened 2021-12-23
Feature/birthday notifications UI

fadwamahmoud:
Second Iteration of the birthday menu feature.
fixes: #266 

- Birthday menu is limited to 6 athletes.
- Birthdays are shown next to Athlete.
- Show All button for the entire list grouped by their birthdays.
- Separate birthday dialog component.

Yet to be implemented:
- UI improvement.
- Badge for number of athlete whose birthdays are on the same week.


eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayDialog.tsx
```diff
@@ -0,0 +1,113 @@
...
+        </Typography>
+        <IconButton className={classes.closeButton} onClick={() => onClose()}>
+          <Close />
+        </IconButton>
+
+        <div className={classes.tableContainer}>
+          {customers.map((customer) => {
+            const customerBirthday =
+              customer.birthday === DateTime.now().toISODate().substring(5)
+                ? t("Today")
```
silviot, line 45: Should be `DateFormat.Today`

eisbuk-admin/src/translations/it.json
```diff
@@ -129,5 +130,12 @@
     "CancelSlot": "Annulla creazione dello slot",
     "ConfirmCreateSlot": "Conferma creazione dello slot",
     "ConfirmEditSlot": "Conferma modifiche allo slot"
+  },
+  "Title": {
+    "UpcomingBirthdays": "Prossimi Compleanni"
```
silviot, line 135: Thanks for taking care of the Italian labels too!

eisbuk-admin/src/enums/translations.ts
```diff
@@ -128,6 +138,7 @@ export enum DateFormat {
   MonthYear = "Date.MonthYear",
   Time = "Date.Time",
   Placeholder = "Date.Placeholder",
+  Today = "Today",
```
silviot, line 141: Is this used somewhere?

eisbuk-admin/src/translations/en.json
```diff
@@ -129,5 +130,11 @@
     "CancelSlot": "Cancel slot creation",
     "ConfirmCreateSlot": "Confirm slot creation",
     "ConfirmEditSlot": "Confirm slot edits"
+  },
+  "Title": {
+    "UpcomingBirthdays": "Upcoming Birthdays"
+  },
+  "MenuButton": {
+    "ShowAll": "Show All"
```
silviot, line 138: This menu button says "show all". I can understand it by reading this file.

But the context is not clear: does this "show all" display on bookings? I can't really tell by just watching here.
I propose we change the position of these labels a bit: we can introduce a new section `BirthdaysMenu` to host the two strings that we need. Then we can move the two existing translations inside it.

Something like this:

```json
{
    "BirthdayMenu": {
        "ShowAll": "Show All",
        "UpcomingBirthdays": "Upcoming Birthdays"
    }
}
```

It's a bit of a tedious change, since it requires updating both translations and the translations enums, but I think it's worth it.

silviot: To clarify: in Italian we might need different translations for `Show All` depending on the gender of the "thing" in `all`. In Italian things have a gender...
So we might need two (potentially more) different `Show All` translatable strings...

eisbuk-admin/src/enums/translations.ts
fadwamahmoud: Yeah, it is used in the birthday menu and dialog but it wasn't working and I hadn't noticed because it was falling back to this value "Today".
It should instead be "Date.Today".

eisbuk-admin/src/translations/en.json
fadwamahmoud: Well, in this case it needs to say `Show All [Birthdays/Items]` so I changed the Italian to "Mostra Tutti". Is this correct?

silviot (approved):

eisbuk-admin/src/translations/en.json
silviot: Yep! It's the kind of difference I meant. In this case _tutti_ is slightly better than _tutto_.

Issue #102, opened 2021-07-02
Add organization settings


silviot:
The organization currently only has one field, `admins` (a list of strings).
It could host other fields to customize the UI and behaviour.
A new route is also needed (`/admin_preferences`) with a form to edit them.

|Field| name| 
|------ | ------ |
|Oganization name| `name`|
|List of admins| `admins`|
|Email address to use as `from` for sent emails | `emailFrom`|
|Name to use when sending emails ( the name part in `Name <email@example.com>`) | `emailNameFrom` |
|Template for reminder emails |`emailTemplate`|
|Caller ID to use when sending out SMSs| `smsFrom`|
|Template for reminder SMSs| `smsTemplate`|
|~SMTP URI to use for outgoing emails~ This will be in the `secrets` collection, with no UI | ~`smtpUri`~|

PR #285, opened 2022-01-02
Feature/sms booking link

### LINKED ISSUE #280: Button to send booking link via SMS
Next to the send-by-email button there should be a button with similar functionality, but to send an SMS

- [x] Create (and test) UI and client side logic
- [x] Implement cloud function in charge of handling the SMS sending itself

ikusteu:
Fixes #280 
The button (and client functionality is created), but there is no https handler (cloud function for SMS sending), hence a PR draft.

### DIFF
--- a/eisbuk-admin/functions/src/https.ts
+++ b/eisbuk-admin/functions/src/https.ts
@@ -12,7 +12,6 @@ import {
   SendEmailErrors,
   SendSMSErrors,
   OrganizationData,
-  OrgSubCollection,
   OrganizationSecrets,
 } from "eisbuk-shared";
 
@@ -96,16 +95,15 @@ export const sendSMS = functions
       );
     }
 
-    const orgRef = admin
-      .firestore()
-      .collection(Collection.Organizations)
-      .doc(organization!);
-
     // get SMS template data
-    const orgSnapshot = await orgRef.get();
-
     const { sender: templateSender, smsUrl } =
-      orgSnapshot.data() as OrganizationData;
+      (
+        await admin
+          .firestore()
+          .collection(Collection.Organizations)
+          .doc(organization!)
+          .get()
+      ).data() || ({} as OrganizationData);
 
     // check template
     if (!smsUrl) {
@@ -116,21 +114,23 @@ export const sendSMS = functions
     }
 
     // get sms secrets
-    const smsSecrets = (
-      await orgRef.collection(OrgSubCollection.Secrets).doc("sms").get()
-    ).data() as OrganizationSecrets["sms"];
+    const { smsAuthToken: authToken } =
+      (
+        await admin
+          .firestore()
+          .collection(Collection.Secrets)
+          .doc(organization!)
+          .get()
+      ).data() || ({} as OrganizationSecrets);
 
-    if (!smsSecrets?.authToken) {
+    if (!authToken) {
       throw new functions.https.HttpsError(
         "not-found",
         SendSMSErrors.NoAuthToken
       );
     }
 
-    const { proto, ...options } = createSMSReqOptions(
-      smsUrl,
-      smsSecrets.authToken
-    );
+    const { proto, ...options } = createSMSReqOptions(smsUrl, authToken);
 
     const data = new TextEncoder().encode(
       JSON.stringify({
--- a/eisbuk-admin/functions/src/https.ts
+++ b/eisbuk-admin/functions/src/https.ts
@@ -96,14 +96,14 @@ export const sendSMS = functions
     }
 
     // get SMS template data
-    const { sender: templateSender, smsUrl } =
-      (
+    const { smsSender: templateSender, smsUrl } =
+      ((
         await admin
           .firestore()
           .collection(Collection.Organizations)
           .doc(organization!)
           .get()
-      ).data() || ({} as OrganizationData);
+      ).data() as OrganizationData) || {};
 
     // check template
     if (!smsUrl) {
--- a/eisbuk-admin/functions/src/https.ts
+++ b/eisbuk-admin/functions/src/https.ts
@@ -143,6 +143,8 @@ export const sendSMS = functions
 
     const res = await postSMS(proto, options, data);
 
+    functions.logger.log("SMS POST request successfull, response:", res);
+
     return { success: true, res };
   });
 
--- a/eisbuk-admin/functions/src/https.ts
+++ b/eisbuk-admin/functions/src/https.ts
@@ -1,14 +1,19 @@
+import https from "https";
 import * as functions from "firebase-functions";
 import admin from "firebase-admin";
 
 import {
   Collection,
   SendMailPayload,
+  SendSMSPayload,
   HTTPErrors,
   SendEmailErrors,
+  SendSMSErrors,
+  OrganizationData,
 } from "eisbuk-shared";
 
 import { checkUser } from "./utils";
+import { StringDecoder } from "string_decoder";
 
 /**
  * Stores email data to `emailQueue` collection, triggering firestore-send-email extension.
@@ -24,7 +29,7 @@ export const sendEmail = functions
       );
     }
 
-    const { organization, ...email } = payload!;
+    const { organization, ...email } = payload;
 
     await checkUser(organization, auth);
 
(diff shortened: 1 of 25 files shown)


ikusteu:
To test out the behaviour (since we don't have admin preferences UI yet), you need to manually add populate the organization document in firestore with `sender` (string) and `smsConfig` (map) of `authToken`: <token from gateway api> and `url`: https://gatewayapi.com/rest/mtsms (without the "?token=<authToken>" part)

eisbuk-admin/functions/src/https.ts
```diff
@@ -53,3 +59,126 @@ export const sendEmail = functions
...
+
+    const { organization, to, message } = payload;
+
+    await checkUser(organization, auth);
+
+    // check payload validity
+    if (!to) {
+      throw new functions.https.HttpsError(
+        "invalid-argument",
+        SendEmailErrors.NoRecipient
```
silviot, line 87: Shouldn't this be `SendSMSErrors.NoRecipient`?

eisbuk-admin/functions/src/https.ts
```diff
@@ -53,3 +59,126 @@ export const sendEmail = functions
...
+      );
+    }
+    if (!message) {
+      throw new functions.https.HttpsError(
+        "invalid-argument",
+        SendEmailErrors.NoMsgBody
```
silviot, line 93: Similarly, Shouldn't this be `SendSMSErrors.NoMsgBody`?

eisbuk-shared/src/enums/errorMessages.ts
```diff
@@ -10,4 +10,10 @@ export enum SendEmailErrors {
   NoMsgBody = "Message body (html) not provided",
   NoSubject = "No email subject specified",
 }
+
+export enum SendSMSErrors {
+  NoMsgBody = "Message body not provided",
```
silviot, line 15: Should we also have a `NoRecipient` here?

eisbuk-admin/src/translations/en.json
```diff
@@ -68,7 +68,9 @@
     "DeleteSlot": "Are you sure you want to delete the",
     "NonReversible": "This action is non-reversible",
     "SendEmailTitle": "Confirm email",
-    "ConfirmEmail": "Are you sure you want to send an email to:"
+    "ConfirmEmail": "Are you sure you want to send an email to:",
```
silviot, line 71: We should not assume a specific structure of the message (like here `fixedText` `customerName``?`).
[We should use interpolation](https://www.i18next.com/translation-function/interpolation) to leave a spot in the translation string that will be filled in when invoking `translate`.

eisbuk-admin/src/components/atoms/CustomerCard/ActionButtons.tsx
```diff
@@ -70,27 +78,76 @@ const ActionButtons: React.FC<ActionButtonProps> = ({
...
+ * @param {string} param.phone (optional) customer's phone
+ * @returns
+ */
+const getDialogPrompt: GetDialogPrompt = ({ method, email, phone }) => {
+  if (!method) return { promptTitle: "", promptBody: "" };
+
+  const promptLookup = {
+    [SendBookingLinkMethod.Email]: {
+      promptTitle: i18n.t(Prompt.SendEmailTitle),
+      promptBody: [i18n.t(Prompt.ConfirmEmail), email, "?"].join(" "),
```
silviot, line 140: After changing the translation string this should be changed to `i18n.t(Prompt.ConfirmEmail, {email})`

eisbuk-shared/src/types/firestore.ts
```diff
@@ -42,6 +42,24 @@ import {
...
+   */
+  smsConfig?: {
+    /**
+     * Full URL (including host and endpoint) of SMS sending service provider.
+     */
+    url?: string;
+    /**
+     * Auth token used to authenticate with SMS sending service provider
+     */
+    authToken?: string;
```
silviot, line 61: I'm a bit concerned with all the places this auth token will go to. I'd rather prefer it staying home, and never leaving.
Organization attributes are currently sent to the client, but the client doesn't really need them.
I think a better approach could be creating a new `secrets` subcollection in `organizations`. This `secrets` collection should not be readable by any user. Just the allmighty backend code that runs in the firebase cloud functions should be able to read it.
I think this should be part of this PR:
* new rules to deny read permission to anyone to the `secrets` subcolleciton
* adjust code to pull data from said collection

The UI to change these values (not to see/read them) should be part of #102. If possible we could set firestore rules to let admin users write to the `secrets` subcollection, but not read from it. If that's tricky/not possible we could have a cloud function that sets these values when invoked by an admin.

silviot (changes requested):

eisbuk-admin/functions/src/https.ts
ikusteu: Noted, will checking now

eisbuk-admin/src/components/atoms/CustomerCard/ActionButtons.tsx
ikusteu: :+1:

ikusteu:
Update on the data model for testing/adding UI updating functionality:
- `sender` (optional `string`) is stored withing organization document (i.e. `organizations/eisbuk-test`) and is to be used as sender for SMS and (comming soon) email
- `smsUrl` (`string`) is a provider url (i.e. `https://gatewayapi.com/rest/mtsms`, without query string) for SMS, also stored in organization document (see above)
- `smsAuthToken` (`string`) is an auth token used to authenticate with SMS service provider and is stored in secrets document for an organization (following the above example, this would be `secrets/eisbuk-test`). This document is write only to the client (in the future) and read only by cloud functions

Additionally, there is `existingSecrets` (`string[]`) entry in organization document (accessible by admin authenticated client) to list all of the existing secrets, without accessing said secrets' values

silviot:
@ikusteu I tried testing this, but ran into some problems.
What I did:
* Deployed functions and firestore rules wih `npx firebase deploy --project eisbuk --only functions` and `npx firebase deploy --project eisbuk --only firestore`.
* [Added secrets configuration](https://console.firebase.google.com/u/0/project/eisbuk/firestore/data/~2Fsecrets~2Figorice.web.app) and [sender/smsUrl](https://console.firebase.google.com/u/0/project/eisbuk/firestore/data/~2Forganizations~2Figorice.web.app).
* Navigated to https://igorice--285-merge-kbmlq7si.web.app/athletes and clicked "Send SMS".

I get this in the cloud functions log:

```
Unhandled error Error: Value for argument "collectionPath" is not a valid resource path. Path must be a non-empty string.
    at Object.validateResourcePath (/workspace/node_modules/@google-cloud/firestore/build/src/path.js:446:15)
    at Firestore.collectiUnhandled error Error: Value for argument "collectionPath" is not a valid resource path. Path must be a non-empty string.
on (/workspace/node_modules/@google-cloud/firestore/build/src/index.js:591:16)
    at /workspace/dist/index.js:1:47756
    at    at Object.validateResourcePath (/workspace/node_modules/@google-cloud/firestore/build/src/path.js:446:15)
 Generator.next (<anonymous>)
    at n (/workspace/dist/index.js:1:1241)
    at Firestore.collection (/workspace/node_modules/@google-cloud/firestore/build/src/index.js:591:16)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)     at /workspace/dist/index.js:1:47756
    at Generator.next (<anonymous>)
    at n (/workspace/dist/index.js:1:1241)
```

I'm not sure if I did some mistake in my testing or if the code needs fixing.

ikusteu:
I think I've managed to reproduce @silviot's issue, by accident, just as, I believe, the issue arose in the first place. I switched from different branch to `feature/sms_bookings_link`, didn't run `yarn` in the root of the project == didn't build the `eisbuk-shared` package. 
This caused the `Collection.Secrets` to resolve to `undefined`, rather than `"secrets"` as `Collection` enum is loaded from (unbuilt) `eisbuk-shared`. This in turn caused an "empty" (=`undefined`) path segment and firestore admin SDK threw an error while checking path vaildity (it being not empty). 
The bug is fixed and both data trigger function and SMS sending functions work correctly (tested on both `eisbuk.web.app` and `igorice.web.app`)

There's that, but I've found some additional bugs that need solving:
- [ ] Success message for SMS sending is an email success message, rather than SMS
- [ ] Phone number isn't correctly interpolated (in `ConfirmDialog`) while translating, nor is email
- [ ] We're getting "stateful update on an unmounted component" error (my suspicion is closing of the `CustomerDialog`, too early, and maybe we should make it not close automatically on send-sms/send-email button press)

ikusteu:
@silviot please take a look and correct the Italian translations in `@/translations/it.ts` (in `Notifications` and `Prompt`). I'm leaving this unmerged so that we can have appropriate translations when it's merged and the PR is more/less standalone (the rest of the work doesn't really depend on it)

silviot (approved):

Issue #278, opened 2021-12-29
Fix firestore subscription date change


ikusteu:
Could be done with #239 

Firestore subscriptions didn't change on date change (a miss-out while initially developing the functionality). The quick fix has been applied -> to resubscribe all collections each time a date is changed. This is obviously sub-optimal in terms of performance and costs.

The fix should be to check each collection subscribed to and resubscribe only the ones that require it.

PR #289, opened 2022-01-06
Fix/ordered attendance customers

### LINKED ISSUE #276: Athlete ordering in attendance view should not change when adding/removing/changing time intervals
Currently the list of athletes that did not book might cchange its ordering when you change the time interval an athlete attended.
They should instead be always sorted by surname.

ikusteu:
Fixes #276 
The fix is made by sorting customers inside selector, so not too optimized, but we can address optimized approach when (and if) needed.


silviot:
Sorting per se is not partucularly "non-optimized". Sorting with comparison has an asymptotic complexity of `n log(n)`.

[In absolute terms](https://www.measurethat.net/Benchmarks/Show/3549/0/javascript-sorting-algorithms) my browser can sort a 1k elements array in microseconds (700 µs at most, when using the most naive of algorythms, the bubble sort, that's mostly useful in education for introductions to sorting algorythms). 50 µs with the inplace quicksort.

There are however some problems with the comparison function in this PR. I'll elaborate with comments on the code.

eisbuk-admin/src/components/atoms/AttendanceCard/AttendanceCard.tsx
```diff
@@ -88,12 +88,16 @@ const AttendanceCard: React.FC<Props> = ({
...
-          !customers.find(({ id: customerId }) => customerId === id)
-      ),
-    [customers]
+      allCustomers
+        .filter(
+          ({ category, id }) =>
+            categories.includes(category) &&
+            !attendedCustomers.find(({ id: customerId }) => customerId === id)
+        )
+        .sort(({ name: a1, surname: a2 }, { name: b1, surname: b2 }) =>
```
silviot, line 97: There are a few problems with the comparison function passed to `sort`.

```javascript
({ name: a1, surname: a2 }, { name: b1, surname: b2 }) =>
    `${a1} ${a2}`.toLowerCase() < `${b1} ${b2}`.toLowerCase() ? -1 : 1
```

* we should sort by last name first, then by first name
* the function needs to allocate memory to create two new strings
* the two new strings need to be lowercased, possibly needing more memory allocated for the lowercase version

This is really not needed at all. I doubt the impact of this exceeds a millisecond, so it would not be a huge problem if this went in, but for the sake of beauty, let's make it better.

I think the algorythm expressed with this pseudocode would yield the same result without allocating new memory. It also provides faster exit from the function, so that if the first letter of the last name is different the first name is not read at all.

* If a2 > b2 return 1
* If a2 < b2 return -1
* // if we're still executing at this point, it means a2 === b2
* if a1 > b1 return 1
* if a1 < b1 return -1
* // If we're at this point it means both `a`s and `b`s are equal
* return 0

This method should also be used for the sorting in `eisbuk-admin/src/store/selectors/attendance.ts`

silviot (changes requested):
The comparison function needs to be amended to be more memory and time efficient.

ikusteu:
The comparison function is fixed (the sorting of the `filtered customers` was moved outside of the component (to `getCustomersList` selector). 
The problem in production, is "flashing" of the attendance page on each date change. This, however, is due to resubscription on each date change (should be optimized), which is out of scope of this PR and is addressed in #278

eisbuk-admin/src/utils/customers.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * A comparison function passed as `Array.sort` callback to sort
+ * an array of customers alphabetically, in an accending order, according to surname
+ * and then the name (if surname is the same)
+ */
+export const compareCustomers = (
+  { name: nm1, surname: sn1 }: Customer,
+  { name: nm2, surname: sn2 }: Customer
+): number => {
+  switch (true) {
```
silviot, line 12: TIL about this pattern

eisbuk-admin/src/utils/customers.ts
```diff
@@ -0,0 +1,28 @@
...
+    case sn1 > sn2:
+      return 1;
+    case sn1 < sn2:
+      return -1;
+    // if we made it this far, the surnames are the same
+    // compare names
+    case nm1 > nm2:
+      return 1;
+    case nm1 < nm2:
+      return 1;
```
silviot, line 23: This should probably be `-1`

silviot (changes requested):
There's a `1` that should be a `-1`. Otherwise looks good.

eisbuk-admin/src/utils/customers.ts
ikusteu: :wink:

silviot (approved):

PR #301, opened 2022-01-17
Feature/firebase subscription followup

### LINKED ISSUE #278: Fix firestore subscription date change

ikusteu:
Fixes #278

A fix to use firestore subscriptions to only subscribe to difference (for date constrained subscriptions) when paginating (using date). The current implementation uses multiple subscriptions to only subscribe to additional range i.e.:
We're subscribed to slots for "2022-01", "2022-02" and "2022-03" and wish to subscribe to "2022-02", "2022-03" and "2022-04". Instead of performing redundant subscriptions (months "2022-02" and "2022-03"), the listener will only subscribe to "2022-04", leaving the initial subscription (first three months) intact and storing the `unsubscribe` function which will unsubscribe all listeners: in this case the first (three months) and the second (one month). This, of course, produces a lot of unnecessary complexity and code (as much as I've tried to keep it as clean as possible), but it works and is very exact.

However, I've found that the pricing would be the same even if we had redundant subscriptions (multiple listeners subscribed to the same document) as each document update will result in one "request" (in terms of firebase billing) and the multiple listener updates would take place locally. Thus, I will update this functionality (probably as a separate npm package) in the future, to use less code, be more readable and offer a lot more flexibility (be generic and project agnostic).


silviot:
I've quickly tried the code here by [clicking on the deployed link in the comment above](https://eisbuk--301-merge-lh92r17x.web.app/) and opening the redux dev tool pane.

I let the app open, then I clicked `Commit` in the redux dev tools, to get rid of all state changes and focus on what actions are dispatched when the user changes the date.

I clicked to change the date by one day. I saw a lot of actions being dispatched. I think it's a huge problem.

This is what I saw:
image

A date change should not trigger an action per user. As things are currently changing date will take more and more time as more users are added. It's not needed nor acceptable.

eisbuk-admin/src/react-redux-firebase/reducer/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    case Action.UpdateLocalDocument:
+      const {
+        collection: collectionToUpdate2,
+        data: updatedDocument,
+        id: documentToUpdate,
+      } = action.payload as FirestoreReducerAction<Action.UpdateLocalDocument>["payload"];
+      return {
+        data: {
+          ...state.data,
+          [collectionToUpdate2]: {
```
silviot, line 67: It looks like the `2` in the variable name is there to differentiate it from a different `collectionToUpdate` in the same block.

I propose to introduce blocks to scope variables when there's a similar need, instead of changing variable names. An opening bracket after the case statement (`case Action.UpdateLocalDocument: {`) and a closing one before the next `case` statement look more readable to me.

silviot: BTW this should already be changed in the branch I'm working on. No action needed.

eisbuk-admin/src/react-redux-firebase/reducer/index.ts
```diff
@@ -0,0 +1,83 @@
...
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { [collectionToDelete]: _coll, ...data } = state.data;
+      // remove the listener for a collection from the store
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { [collectionToDelete]: _listener, ...listeners } = state.listeners;
+      // return state with listener for collection and the correspoding data removed
+      return { data, listeners };
+
+    case Action.UpdateLocalDocuments:
+      const { collection: collectionToUpdate1, data: updatedCollection1 } =
```
silviot, line 44: I believe it would be better to use the variable name `collectionToUpdate` enclosing the whole code for this `case` in a block.

silviot (approved):
LGTM except for the numbered variables that would benefit from losing their number IMO.

PR #303, opened 2022-01-17
Sort intervals by start-time and duration-reverse

silviot:


fadwamahmoud (approved):

eisbuk-admin/src/__tests__/utils.test.ts
```diff
@@ -79,4 +84,16 @@ describe("Date utils", () => {
...
+  describe("`comparePeriods` function", () => {
+    test("should return earlier periods first", () => {
+      expect(comparePeriods("13:00-13:30", "13:30-14:00")).toEqual(-1);
+      expect(comparePeriods("13:00-13:30", "12:30-13:00")).toEqual(1);
+    });
+
+    test("should return longer periods (starting at the same time) first", () => {
+      expect(comparePeriods("13:00-14:30", "13:00-14:00")).toEqual(-1);
+      expect(comparePeriods("13:00-13:30", "13:00-14:00")).toEqual(1);
+    });
```
ikusteu, line 97: I understand this is a very simple functionality, but I would rather test them as close to production as possible. In this case, I would do something like this: 
```typescript
expect(["13:00-13:30", "13:30-14:00"].sort(comparePeriods)).toEqual([
        "13:00-13:30",
        "13:30-14:00",
      ]);
```
For the following reasons: 
- this is more on the side of "testing behavior rather than implementation"
- I never remember the JS sort rule (the -1, 1 thing), as simple and logical as it is, it just doesn't stick with me and I need to think it through each time

ikusteu: Here's a solution copy/paste to save you some typing in case you decide to go this route
```typescript
describe("`comparePeriods` function", () => {
    test("should return earlier periods first", () => {
      expect(["13:00-13:30", "13:30-14:00"].sort(comparePeriods)).toEqual([
        "13:00-13:30",
        "13:30-14:00",
      ]);
      expect(["13:00-13:30", "12:30-13:00"].sort(comparePeriods)).toEqual([
        "12:30-13:00",
        "13:00-13:30",
      ]);
    });

    test("should return longer periods (starting at the same time) first", () => {
      expect(["13:00-14:30", "13:00-14:00"].sort(comparePeriods)).toEqual([
        "13:00-14:30",
        "13:00-14:00",
      ]);
      expect(["13:00-13:30", "13:00-14:00"].sort(comparePeriods)).toEqual([
        "13:00-14:00",
        "13:00-13:30",
      ]);
    });
  });
```

silviot: I tried to use the `comparePeriods` function directly, but it's true that it's always passed on to `sort`.
I understand the unease with the `-1`/`1` results. I try to remember that they point to where the smallest value is. I then imagine the numbers line, with `-1`, `0` and `1` in that order from lest to right. A result `1` means the smallest item is on the right. A result of `-1` means that it's on the left.
I'll apply the change anyway. Thanks!

ikusteu (approved):

PR #308, opened 2022-01-19
Dispatch a single action to update multiple records

silviot:
The linting fails with

```
src/react-redux-firebase/reducer/__tests__/firestoreReducer.test.ts:98:59 - error TS2345: Argument of type '{ type: Action.UpdateLocalDocuments; payload: { collection: OrgSubCollection.Customers; records: { data: Customer; id: string; }[]; }; }' is not assignable to parameter of type 'FirestoreReducerAction<FirestoreAction>'.
  Types of property 'payload' are incompatible.
    Type '{ collection: OrgSubCollection.Customers; records: { data: Customer; id: string; }[]; }' is not assignable to type 'CollectionSubscription | { collection: CollectionSubscription; records: { data: never; id: string; }[]; } | { collection: CollectionSubscription; ids: string[]; } | { ...; }'.
      Type '{ collection: OrgSubCollection.Customers; records: { data: Customer; id: string; }[]; }' is not assignable to type '{ collection: CollectionSubscription; records: { data: never; id: string; }[]; }'.
        Types of property 'records' are incompatible.
          Type '{ data: Customer; id: string; }[]' is not assignable to type '{ data: never; id: string; }[]'.
            Type '{ data: Customer; id: string; }' is not assignable to type '{ data: never; id: string; }'.
              Types of property 'data' are incompatible.
                Type 'Customer' is not assignable to type 'never'.

98       const updatedState = firestoreReducer(initialState, updateAction);
```

@ikusteu Can you have a look at that? I believe it should be a very quick one for you, and I didn't investigate the `never` type (as I understand it's a temporary solution I didn't put much effort into looking into it).

### DIFF
--- a/eisbuk-admin/src/enums/store.ts
+++ b/eisbuk-admin/src/enums/store.ts
@@ -22,9 +22,8 @@ export enum Action {
   Logout = "@@Eisbuk/LOGOUT",
 
   UpdateFirestoreListener = "@@Eisbuk/UPDATE_FIRESTORE_LISTENER",
-  UpdateLocalCollection = "@@Eisbuk/UPDATE_LOCAL_COLLECTION",
-  UpdateLocalDocument = "@@Eisbuk/UPDATE_LOCAL_DOCUMENT",
-  DeleteLocalDocument = "@@Eisbuk/DELETE_LOCAL_DOCUMENT",
+  UpdateLocalDocuments = "@@Eisbuk/UPDATE_LOCAL_DOCUMENTS",
+  DeleteLocalDocuments = "@@Eisbuk/DELETE_LOCAL_DOCUMENTS",
   DeleteFirestoreListener = "@@Eisbuk/DELETE_FIRESTORE_LISTENER",
 }
 /**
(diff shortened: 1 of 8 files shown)


ikusteu:
Ah right...sorry, forgot about that...it's a result of complex type inheritance (gone wrong). 
FYI type never is this:
```typescript
const throwErr = (): never => {
  throw new Error()
}
```
So this type throws an error and never returns...and in the case at hand...it's just a type inheritance gone wrong

ikusteu:
I'm thinking of rejecting this PR and just fixing the original implementation with this in mind. The proposed solutions is this:
To use the `updateLocalColl` (which is there to update the local collection (`customers`, `attendances`, etc.) with `{merge: true}`. This will have the same behavior as updating multiple documents, only: instead of adding each document to an array and passing `{[id], [data]}`, and then unwrapping that array and creating an object looking like this `{[id]: data}`, to create `{[id]: data}` payload (as is implemented in the first place) and just adding this to updated collection state (in local store). I other words, the reducer would receive the data and update it as is (without the need to iterate over an array). And keep in mind the mental model of still having the `deleteLocalDocument` action to delete the documents which weren't in the update (=are deleted)
In conclusion, I think we would achieve the same thing, only with less code and less iterations. Please dispute me on semantics of `updateLocalColl`...or the proposed solution for that matter, but it seems pretty straightforward to me.

silviot:
@ikusteu I have no objections to this plan: my concern is about needlessly splitting one "unit of work" into many. As long as that does not happen I'm fine with it.

PR #309, opened 2022-01-19
Fix user discrepency of multiple views

### LINKED ISSUE #299: When two users use the "attendance view they don't have a consistent view!
One of the users might have rows end up stuck in a "disabled" state.
Try to open two browser windows on the same attendance date and marking athletes as present/absent.

ikusteu:
Fixes #299


silviot (approved):
The fix looks good to me.

Regarding the test, I think an easier to understand way would be write a cypress test where we use the firestore API to change a record while the browser window is open.

Merging as is since it fixes an annoying bug.

PR #316, opened 2022-01-20
Feature/booking duration UI

### LINKED ISSUE #298: Clearly state duration on booking form
When users book a slot they can infer the duration by doing the math.

Math is hard.

We should task a computer to do the math.

We should display the slot duration in text (`1H`, `1 1/2H`, `2H`) and maybe with icons:
* one hour hour
* one hour and a half hour half-hour
* two hours hour hour)

These are the SVG files of the above:
* noun-hour-766839

*  noun-half-hour-766750

ikusteu:
Fixes #298 
This is ready to go in as-is, but doesn't need to. This is just a visual suggestion (instead of proposed SVGs). I forgot to comment that the SVGs look more like the sun (rather than the clock) to me, so the first thing that comes to mind is day rather than hour...please dispute this if you thing it needs more UI work/SVG integration


silviot:
@ikusteu I think the text needs to be larger. There seems to be space next to the time in the top line. Maybe it makes sense to color code the three possible values (and the impossible one) to ease eye scanning.

silviot (approved):

PR #317, opened 2022-01-20
Add visual difference between categories and type tags on AttendanceCard

### LINKED ISSUE #137: Set up some visual difference between slot type and category in attendance view
It might be a good idea to separate slot types and categories for slots in bookings view and otherwise. Right now, they're just written out one after another. My suggestion is to do something like GitHub's labels (colored tags).

ikusteu:
Fixes #137
A simple visual distinction between tags and categories in the `AttendanceCard`

### DIFF
--- a/eisbuk-admin/src/__testData__/slots.ts
+++ b/eisbuk-admin/src/__testData__/slots.ts
@@ -55,7 +55,7 @@ export const baseSlot: SlotInterface = {
   date: testDate,
   id: "base-slot-id",
   type: SlotType.Ice,
-  categories: [Category.Competitive, Category.PreCompetitive],
+  categories: [Category.Competitive, Category.Adults],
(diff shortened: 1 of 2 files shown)


silviot:
The change looks good, but tests need to be fixed.
Their failure uncovered a couple of problems:
* [the `Test Report`](https://github.com/eisbuk/EisBuk/blob/master/.github/workflows/build.yml#L102-L108) step might fail with `Error: Validation Failed: {"resource":"CheckRun","code":"invalid","field":"annotations"}`
* the verbosity of our current tests means that a poor soul needs to sift through 3731 output lines to find out what went wrong

Feel free to give either or both a thought (like reducing the test logs by a few hundred lines would be already beneficial).

silviot (approved):

PR #318, opened 2022-01-20
Rerender birthday menu on midnight

fadwamahmoud:


eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.tsx
```diff
@@ -37,6 +38,17 @@ const BirthdayMenu: React.FC<Props> = ({ customers, onClickShowAll }) => {
     setBirthdaysAnchorEl(null);
   };
 
+  const calculateTimeDiff =
```
silviot, line 41: `calculateTimeDiff` evokes a function, since it starts with a verb. A nou would be more appropriate, to describe the value contained here. `millisToMidnight` IMO would be more descriptive.

eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.tsx
```diff
@@ -37,6 +38,17 @@ const BirthdayMenu: React.FC<Props> = ({ customers, onClickShowAll }) => {
...
+    DateTime.now().plus({ day: 1 }).startOf("day").toMillis() -
+    DateTime.now().toMillis();
+  const [timeDiff, setTimeDiff] = useState(calculateTimeDiff);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setTimeDiff(calculateTimeDiff);
```
silviot, line 48: Actually, `setTimeDiff would work better as a funtion. Here it needs to be invoked.

eisbuk-admin/src/components/atoms/BirthdayMenu/BirthdayMenu.tsx
```diff
@@ -37,6 +38,17 @@ const BirthdayMenu: React.FC<Props> = ({ customers, onClickShowAll }) => {
...
+    }, timeDiff);
+    return clearTimeout();
```
silviot, line 50: `setTimeout` returns an id. It needs to be captured and used in the _function_ returned here. The function should invoke `clearTimeout`: we don't want to call it now - we want to return a function that will clear it when invoked (i.e. when/if the component is unmounted).

eisbuk-admin/cypress/integration/birthday_menu_spec.ts
```diff
@@ -0,0 +1,48 @@
...
+import { saul } from "@/__testData__/customers";
+import { __birthdayMenu__ } from "@/__testData__/testIds";
+import { DateTime } from "luxon";
+import { __storybookDate__ as __staticTestDate__ } from "@/lib/constants";
+
+describe("birthday badge", () => {
+  beforeEach(() => {
+    cy.initAdminApp();
+    cy.visit(PrivateRoutes.Athletes);
+    cy.clock(DateTime.fromISO(__staticTestDate__).toMillis());
```
silviot, line 11: Is this necessary?

ikusteu (approved):

ikusteu (approved):

PR #320, opened 2022-01-21
Feature/cypress firebase

ikusteu:
Developed a functionality to send filenames from cypress tests back to cypress node process. The filenames are used to read json files from `cypress/__testData__` and update emulated firestore with data from said files


eisbuk-admin/cypress/plugins/firebasePlugin/updateFirestore/adminDb.ts
```diff
@@ -0,0 +1,6 @@
+import admin from "firebase-admin";
+
+process.env["FIRESTORE_EMULATOR_HOST"] = "localhost:8080";
```
silviot, line 3: I think it would be better to set this value when invoking the tests. Or at least leave an already set value there.
Something like:

```typescript
process.env["FIRESTORE_EMULATOR_HOST"] = process.env["FIRESTORE_EMULATOR_HOST"] || "localhost:8080";
```

eisbuk-admin/cypress/plugins/firebasePlugin/updateFirestore/index.ts
```diff
@@ -0,0 +1,26 @@
...
+interface Payload {
+  organization: string;
+  files: string[];
+}
+
+const handleFirestoreUpdate: TaskHandler<Payload> = async ({
+  organization,
+  files,
+}) => {
+  const testFilesDir = path.join(__dirname, "..", "..", "..", "__testData__");
```
silviot, line 17: Thanks for not using the current working directory! ;-)

eisbuk-admin/cypress/plugins/index.ts
```diff
@@ -21,14 +18,17 @@ const {
  * @type {Cypress.PluginConfig}
  */
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-module.exports = (on) => {
-  on("before:run", async (details) => {
-    console.log("override before:run");
-    await beforeRunHook(details);
-  });
+module.exports = (on: Cypress.PluginEvents) => {
```
silviot, line 21: Did the mochawesome reporter interfere with this? Can it be re-enabled?

eisbuk-admin/cypress/plugins/firebasePlugin/updateFirestore/readTestData.ts
```diff
@@ -0,0 +1,66 @@
...
+const readTestData = async (
+  dir: string,
+  files: string[]
+): Promise<FirestoreDataUpdate> => {
+  // read and parse the files
+  const parsedFiles = await Promise.all(
+    files.map((fName) => {
+      // trim ".json" (if included in the filename)
+      // and add ".json" at the end (to ensure the files has only one ".json" extension)
+      const jsonFName = fName.replace(".json", "") + ".json";
```
silviot, line 24: Is it necessary to support two different ways to refer to the same file?
If it really is, this implementation causes a problem with

```javascript
"my-files.json/aaa.json".replace(
  ".json", "") + ".json" === 'my-files/aaa.json.json'
```

To fix this we can specify we need `.json` to be at the end of the string:

```javascript
"my-files.json/aaa.json".replace(
  /\.json$/, "") + ".json" === 'my-files.json/aaa.json'
```
or no automatic appending of `.json` (my favourite).

eisbuk-admin/cypress/plugins/firebasePlugin/updateFirestore/index.ts
ikusteu: Haha, np...this will also be read from the config.

eisbuk-admin/cypress/plugins/index.ts
ikusteu: Not at all...it just produced some errors when using `cypress run` (since it was unreachable/not defined in that environment) so I commented it out. Aand later on, I guess I deleted it by mistake (since it was commented out). I'll go ahead and reinstantiate that.

eisbuk-admin/cypress/plugins/firebasePlugin/updateFirestore/readTestData.ts
ikusteu: Right...I didn't really account for that scenario...and ok, I'll do it without the name cleanup..I guess there are a few equally "right" ways to go about it and if you prefer that...I don't have a preference.

silviot (approved):

PR #327, opened 2022-01-27
Fix week logic in slot view and copy/paste slot week

### LINKED ISSUE #323: Copy week quirk for weeks that span two months: 5 pts
If you "copy" a week that begins in a month and ends on another month only the slots in the earlier month will be selected

### LINKED ISSUE #324: When a week is copied and the date is changed to a new week the "select day" behaviour should not be there: 3 pts
If you copy a week, (and turn all its slots yellow) and navigate to the following week, you can still select/deselect slots. It should only be possible in the current week.

Probably best tackled together with #245

silviot:
This fixes #324
Also fixes #323 which was wrongly described. The issue there is not with the copy (that does indeed copy from Monday to Sunday) but with the week display for slots: it was starting at the currently set day instead of a Monday.

### DIFF
--- a/eisbuk-admin/src/store/selectors/slots.ts
+++ b/eisbuk-admin/src/store/selectors/slots.ts
@@ -12,58 +12,58 @@ import { LocalStore } from "@/types/store";
  * @param startDate start of timeframe
  * @returns created selector
  */
-export const getSlotsForCustomer = (
-  category: Category,
-  timeframe: "week" | "month",
-  startDate: DateTime
+export const getSlotsForCustomer =
+  (category: Category, timeframe: "week" | "month", startDate: DateTime) =>
   // eslint-disable-next-line consistent-return
-) => (state: LocalStore): SlotsByDay => {
-  const allSlotsInStore = state.firestore.data?.slotsByDay;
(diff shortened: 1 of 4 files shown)


ikusteu (approved):

eisbuk-admin/src/store/selectors/slots.ts
```diff
@@ -12,58 +12,58 @@ import { LocalStore } from "@/types/store";
...
-
-        const slotKeys = Object.keys(slotsDay);
-        const newDay = slotKeys.reduce((acc, slotId) => {
-          const { categories } = slotsDay[slotId];
-          return categories.includes(category)
-            ? { ...acc, [slotId]: slotsDay[slotId] }
+  (state: LocalStore): SlotsByDay => {
+    const allSlotsInStore = state.firestore.data?.slotsByDay;
+
+    // return early if no slots in store
```
silviot, line 21: I think our editors are playing tug of wars with 2 spaces/4 spaces indentation.
It gets in the way of reviews, so it would be nice to find out why. A first step should be settle on what we want. My vote goes with my editor (2 spaces).

PR #328, opened 2022-01-28
Feature/organization settings

### LINKED ISSUE #102: Add organization settings

fadwamahmoud:
Fixes #102


eisbuk-admin/src/components/atoms/OrganizationSettings/OrganizationSettings.tsx
```diff
@@ -0,0 +1,396 @@
...
+                <div className={classes.fieldSection}>
+                  <Field
+                    label="SMTP Server"
+                    name="SMTPServer"
+                    className={classes.field}
+                    as={TextField}
+                    aria-label="SMTPServer"
+                    placeholder="SMTP server here"
+                    variant="outlined"
+                    value={organization.SMTPServer}
```
ikusteu, line 71: This `value` will stay static. You either want to get values from `Formik`'s render function `({errors, isSubmitting, values}) => <Field value={values.<fieldName>}>`. This way you would need to manually set value for each field using `onChange` handler.
Better approach would be to avoid this complexity and let the `Field` component control value displaying and updateing as it is programmed to do just that, the only input it needs to connect to Formik context is the right `name` prop (as we've discussed)

eisbuk-admin/src/components/atoms/OrganizationSettings/OrganizationSettings.tsx
```diff
@@ -0,0 +1,396 @@
...
+                    label="SMTP Port"
+                    name="SMTPPort"
+                    className={classes.field}
+                    as={TextField}
+                    aria-label="SMTPPort"
+                    placeholder="SMTP port here"
+                    variant="outlined"
+                    value={organization.SMTPPort}
+                    disabled={!enableEdit}
+                  />
```
ikusteu, line 84: There is a lot of repetition (=unnecessary code) here. I see the way to reduce code and improve readability by storing each field params in an array and then render the same "template" for each field, example:
```typescript
const smtpFields: Partial<FieldProps>[] = [
  {
    // with the `name` you need to think of variable naming consistenty and conventions: use camelCase
    name: "smtpHost", 
    label: "SMTP Host"
  },
  {
    name: "smtpPort"
    label: "SMPT Port"
  }
]

return smtpFields.map((field) => (
  <
    {...field} // name and label
    // standard values shared across all five fields
    className={classes.field}
    as={TextField}
    aria-label={field.label}
    placeholder={field.label} // This might be unnecessary, if not you can just add `placeholder` in smptFields structure
    variant="outlined"
    // value={organization.SMTPPort} value is unnecessary for reasons explained above
    disabled={!enableEdit}
  />
)
```

silviot (changes requested):
A couple of changes are needed:

- [x] The `smsUrl` field should be removed
- [x] Occurrencies of `smsSender` in `functions/src/https.ts` should be replaced with `smsFrom`
- [x] The _Oganization name_ `name` field is missing

eisbuk-admin/src/components/atoms/OrganizationSettings/OrganizationSettings.tsx
```diff
@@ -0,0 +1,397 @@
...
+                      aria-label={field.label}
+                      variant={enableEdit ? "outlined" : "filled"}
+                      disabled={!enableEdit}
+                      multiline={
+                        field.name?.includes("Template") ? true : false
+                      }
+                      rows={field.name?.includes("Template") ? "4" : "1"}
+                      value={values[`${field.name}`] || ""}
+                    />
+                  ))}
```
ikusteu, line 137: There are two identical `Array.map()` calls: for sms and email fields. It would be better (for code redundancy and readability) to create a `renderTextfields` function (or whatever you want to name it) which would accept a parameter of `typeof emailFields` and return mapped fields (a simple copy paste of the selected code should do)

eisbuk-admin/src/components/atoms/OrganizationSettings/OrganizationSettings.tsx
```diff
@@ -0,0 +1,397 @@
...
+    dispatch(updateOrganization(orgData, organization.organizationName || ""));
+    setEnableEdit(false);
+  };
+
+  const handleEnableEdit = (resetForm: () => void) => {
+    if (enableEdit) resetForm();
+    setEnableEdit(!enableEdit);
+  };
+
+  const emailFields: Partial<FieldProps>[] = [
```
ikusteu, line 56: I would maybe make a type alias as you would use it throughout: for both `emailFields` and `smsFields` as well as mapping function declaration.

eisbuk-admin/src/components/atoms/OrganizationSettings/OrganizationSettings.tsx
```diff
@@ -0,0 +1,397 @@
...
+                          : classes.field
+                      }
+                      as={TextField}
+                      aria-label={field.label}
+                      variant={enableEdit ? "outlined" : "filled"}
+                      disabled={!enableEdit}
+                      multiline={
+                        field.name?.includes("Template") ? true : false
+                      }
+                      rows={field.name?.includes("Template") ? "4" : "1"}
```
ikusteu, line 173: Instead of `name.includes("Template")` check (which is repeated) you can simply add `multiline: true` in the FieldProps object (of respective multiline field). And then check that flag for the latter condition: `rows={field.multiline ? "4" : "1"}`

ikusteu: Example
```typescript
{
      name: "emailTemplate",
      label: "Email Template",
      multiline: true
}
```

eisbuk-admin/src/components/atoms/OrganizationSettings/OrganizationSettings.tsx
```diff
@@ -0,0 +1,396 @@
...
+          <Button
+            onClick={addAdmin}
+            color="primary"
+            variant="contained"
+            className={classes.addAdminButton}
+          >
+            Add
+          </Button>
+        </div>
+      )}
```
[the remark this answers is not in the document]
ikusteu: I guess you didn't see the comment above ^^^ as it says "pending" for me. I hope you see it now as the proposition still stands. I can help out with that if you want me to.

eisbuk-admin/src/components/atoms/OrganizationSettings/OrganizationSettings.tsx
```diff
@@ -0,0 +1,397 @@
...
+    display: "flex",
+    // marginTop: theme.spacing(2),
+    marginBottom: theme.spacing(1),
+    marginRight: theme.spacing(1),
+    width: "12.5rem",
+  },
+  disabledAdminFieldGroup: {
+    display: "flex",
+    // marginTop: theme.spacing(2),
+    marginBottom: theme.spacing(1),
```
ikusteu, line 338: Remove the commented out properties if they're not to be used (and they don't seem needed)

ikusteu (commented):
I left "a couple" of comments :joy: . Additionally, `OrganizationSettings` could be written as part of `@/pages/admin_preferences/index.ts`. 
It kinda makes sense to do it this way (`OrganizationSettings` component), for storybook previews. But this was the reason I proposed (at a meeting a few days back) to make a field rendering function (`emailFields`/`smsFields`) into a separate component. That way you would reduce the code, follow the "atomic design" principles, have a storybook preview for the component, not the page and have the page file contain all the necessary code.

Please feel free to dispute anything I wrote as well as ping for explanation/help if you need any.

eisbuk-admin/src/components/atoms/OrganizationSettings/OrganizationSettings.tsx
ikusteu: Additionally, I still stand by my proposal that it would be best to make this into a full component, rather than a function.

eisbuk-admin/src/translations/it.json
```diff
@@ -60,7 +60,9 @@
     "InvalidTime": "Formato dell'ora non valido (\"HH:mm\")",
     "InvalidDate": "Formato della data non valido (\"dd/mm/yyyy\")",
     "InvalidPhone": "Inserire un numero di telefono valido, completo di prefisso internazionale (+39 per l'Italia)",
-    "TimeMismatch": "L'ora di inizio è maggiore dell'ora di fine"
+    "TimeMismatch": "L'ora di inizio è maggiore dell'ora di fine",
+    "InvalidSmsFrom": "Formato non valido: Solo lettere o numeri e nessun simbolo/spazio",
```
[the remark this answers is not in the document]
fadwamahmoud: i don't get this one

eisbuk-admin/src/components/atoms/OrganizationSettings/OrganizationSettings.tsx
fadwamahmoud: Actually, I don't remember why I used `Partial` but I don't need to anymore so I'll just leave it as `FieldProps[]`

eisbuk-admin/src/translations/it.json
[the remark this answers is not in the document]
ikusteu: Sms sender doesn't need to be a number (it can be `IgorIce`) for instance. Like...when I get a 2FA SMS from Binance, It says from: `Binance` (not the phone number)

eisbuk-admin/src/AppContent.tsx
```diff
@@ -25,6 +25,7 @@ import usePaginateFirestore from "@/react-redux-firebase/hooks/usePaginateFirest
 import useFirestoreSubscribe from "@/react-redux-firebase/hooks/useFirestoreSubscribe";
 
 import { getIsAuthLoaded, getIsAuthEmpty } from "@/store/selectors/auth";
+import AdminPreferencesPage from "./pages/admin_preferences";
 import { DateTime } from "luxon";
```
ikusteu, line 29: Plesase, just group the imports (so that the file is clean). `AdminPreferencesPage` with the page imports and `luxon` with 3rd party libraries (at the top most group),. even though that might get removed (`luxon` import) when you rebase, as it was there temporarily and was removed by one of the latest branches.

ikusteu (approved):

PR #347, opened 2022-02-05
Feature/bookings due date

### LINKED ISSUE #338: Limit customer's booking to a particular date
This is the first part of #326 
Create logic to check if customer can book a particular slot and disable interval cards if not (both client side and in firestore rules):
- admins can always update bookings
- the checking logic should be a "single source of truth": should be handled by a single function accepting `customer`, `slot` (for booking), `currentDate` and return boolean
- for starters set the deadline for bookings 4 days (starting at 00:00 on the 4th day) before the end of the month, should be editable in the future

### LINKED ISSUE #339: Notify customers when the end of booking period approaches and add a button for customer to save their bookings
The second part of #326 

- [ ] Create `CustomerNorification` UI component showing a message that the booking period is close to expiration in `customer_area`

*note: for flexibility, the "save" button should not appear before booking expiration date minus 2(3) days (this saves the hassle of customers calling admins to "unfreeze" them and simplifies code solution), this is actually trivial as "save" button will be part of `CustomerNotification`, hidden before close-to-expiration period*

### LINKED ISSUE #340: Admins should be able to extend booking period to certain customers
Third part of #326 
- [x] Update booking availability checking logic to check if customer has `extendedDate` (belonging to the current month) and enable booking if the date hasn't passed
- [x] Create UI logic for admin to allow `extendedDate` (should be an action button in `CustomerCard`), this functionality should also "unfreeze" saved bookings for the current month (this should happen under the hood)
- [x] Create different message in `CustomerNotification` during the extended period
- [x] Add a button in `CustomerNotification` to "save" bookings for the current (in effect, upcoming) month, saving the bookings for the month (if booked) or freezing them (if intentionally none booked) silencing the notification:
- should feature confirm dialog

### LINKED ISSUE #346: Create cypress tests for booking flow
I've tested the booking behavior relevant to my ticket (#338), but we should provide tests for full booking flow

ikusteu:
1. Fixes #338 
2. Fixes #339 : 
- requires italian translations for countdown message, hope @silviot can chime in on that
- countdown should probably be restyled a bit, but that's polishing 
3. Fixes #340
4. Fixes #346


ikusteu:
@silviot please provide the translations for `ExtendedBookingDate`, `ExtendBookingDateTitle`, `ExtendBookingDateBody`, `FinalizeBookingsTitle`, `ConfirmFinalizeBookings`, `FinalizeBookings`, `ExtendBookingDate` and `BookingDateExtended` and check the translations for `BookingsCountdownMessage` fields.

eisbuk-admin/src/hooks/useCountdown.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  const now = DateTime.fromMillis(Date.now());
+  const {
+    days: rawDays,
+    hours: rawHours,
+    minutes: rawMinutes,
+    seconds: rawSeconds,
+  } = countdownDate.diff(now, ["days", "hours", "minutes", "seconds"]);
+
+  // turn the numbers into "01", "23", etc strings
```
silviot, line 70: As a user I would prefer seeing `3 days` instead of `03 days`.
I'm also quite hardwired to read a leading `0` as meaning that the following number is in octal (and `031 + 027 == 060` is true in Javascirpt ;-)).
Can we change this to not include a leading zero (and simplify the code a little bit) or am I missing something?

silviot:
I did push some translations (in 007bde93f3a52a57786be6edc7198d60a09f81da) that were lost.
I will cherry-pick that commit, but I'd like to understand why that happened to reduce the likelyness of it happening again.
Could it be that at some poing `push --force-with-lease` did not work and you surrender to the temptation of a `push --force`? If that's the case the solution could have been running a `git pull --rebase` _before_ doing the rebase work.

ikusteu:
Very unlikely...I would also want to know why it happened...I never (like, not once) did `--force` after you told me about `--force-with-lease` and I always do `git pull --rebase origin master` (or whichever branch im pulling for rebase) when I want to make them mergeable

eisbuk-admin/src/hooks/useCountdown.ts
```diff
@@ -0,0 +1,76 @@
...
+  const resetTimeout = () => {
+    if (timeout.current) {
+      clearTimeout(timeout.current);
+      timeout.current = null;
+    }
+  };
+
+  // control 'tick' each second using setTimeout
+  useEffect(() => {
+    const hourInMillis = 1000;
```
silviot, line 35: I think we should trigger this evary hour, instead of every second. I guess the variable name was updated but its value still needs to be changed to 3.6M.

silviot:
As an admin, I extended an athlete deadline to feb 22nd. I saw no effect on the athlete bookings views.

The `extendedDate` value made its way to the `bookings` collection, so I don't think it's a problem with the trigger.

eisbuk-admin/src/hooks/useCountdown.ts
ikusteu: Right...I wanted to preserve the general functionality...will make it so that you can specify the tick length (`"hour"`, `"minute"`, `"second"`). I don't think it makes sense to leave it for day.

ikusteu:
Pushed requested updates

silviot (approved):

PR #348, opened 2022-02-06
Fix sms sending + cleanup

silviot:
The sender for SMSs can never be longer than 11 characters.
These changes truncate it before sending the request.
A couple of files that were useful during development but are no longer relevant were also removed.
The code is already deployed to production.


silviot:
The pre-commit hook was happy with this code, so it's probably missing some cases.
I'm almost sure @ikusteu told me already and opened a ticket about it, but I can't find it.

ikusteu:
It's possible that your changes affected something that wasn't staged and (for efficiency's sake) pre-commit lints only the staged files. Let me check

ikusteu (approved):
LGTM, I've only fixed the type assertion to be a type declaration on a function rather than type assertion (`Promise<SMSResponse>`)

PR #362, opened 2022-02-13
Check SMS delivery status 

### LINKED ISSUE #361: Add check to SMS sending function
When sending SMS, we get a success message if the POST request to `gatewayapi` was successful, but fails silently if there is an issue provider side (insufficient funds, non-whitelisted URL, etc.) We should add a check to see if the message was `ACCEPTED` or `DELIVERED` and return success/error based on that state.

ikusteu:
Fixes #361

### DIFF
--- a/eisbuk-admin/functions/src/https.ts
+++ b/eisbuk-admin/functions/src/https.ts
@@ -1,8 +1,6 @@
-import https from "https";
-import http from "http";
 import * as functions from "firebase-functions";
 import admin from "firebase-admin";
-import { StringDecoder } from "string_decoder";
+import pRetry from "p-retry";
 
 import {
   Collection,
@@ -15,7 +13,7 @@ import {
   OrganizationSecrets,
 } from "eisbuk-shared";
 
-import { checkUser, createSMSReqOptions } from "./utils";
+import { checkUser, createSMSReqOptions, sendRequest } from "./utils";
 
 /**
  * Stores email data to `emailQueue` collection, triggering firestore-send-email extension.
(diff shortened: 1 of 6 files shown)


silviot (approved):

eisbuk-admin/functions/src/https.ts
```diff
@@ -61,104 +59,127 @@ export const sendEmail = functions
...
+          SendSMSErrors.SendingFailed,
+          res
+        );
+      }
+
+      const smsId = res.ids[0];
+
+      const [smsOk, status, errorMessage] = await pRetry(
+        () => checkSMS(smsId, authToken),
+        { maxRetryTime: 5000 }
```
silviot, line 131: I think we should use something else to achieve our goals.

Our goal is to make sure that, no matter what, our function will return in less than X seconds. Imagine, for instance, that the remote service is taking 60 seconds to process requests.

`maxRetryTime` [comes into play when it's time to retry something](https://github.com/tim-kos/node-retry/commit/339a30a8b957a88cd68a8e14bbcf7462b340b4ea). So in the example above, if the remote service always takes 60 seconds, this invocation will take 60 seconds, and not retry a second time (since we exceeded the five seconds allowance `maxRetryTime` gave us.

I think we need something like [`p-timeout`](https://www.npmjs.com/package/p-timeout) and/or the firebase [`timeoutSeconds`](https://firebase.google.com/docs/functions/manage-functions#set_timeout_and_memory_allocation) option to pass to `runWith`.

PR #368, opened 2022-02-13
Remove unnecessary validations from cloud functions

### LINKED ISSUE #363: Reduce unnecessary cloud function error handling
There are some explicit error handlings which can be done implicitly, a lot of those will fail at runtime if invalid payload. We should reduce the payload checks only for errors which would cause a silent fail.

ikusteu:
Fixes #363

### DIFF
--- a/eisbuk-admin/functions/src/https.ts
+++ b/eisbuk-admin/functions/src/https.ts
@@ -6,8 +6,7 @@ import {
   Collection,
   SendMailPayload,
   SendSMSPayload,
-  HTTPErrors,
-  SendEmailErrors,
+  HTTPSErrors,
   SendSMSErrors,
   OrganizationData,
   OrganizationSecrets,
@@ -15,34 +14,32 @@ import {
 
 import { checkUser, createSMSReqOptions, sendRequest } from "./utils";
 
-/**
- * An object of key-value (string-string) pairs
- * where the key is a required field and value is an error message to
- * display if field not present
- */
-interface RequiredFieldsValidation {
-  [field: string]: string;
-}
-
 const checkRequiredFields = (
   payload: Record<string, any> | undefined | null,
-  requiredFields: RequiredFieldsValidation
+  requiredFields: string[]
 ) => {
   if (!payload) {
     throw new functions.https.HttpsError(
       "invalid-argument",
-      HTTPErrors.NoPayload
+      HTTPSErrors.NoPayload
     );
   }
 
+  const missingFields: string[] = [];
+
   Object.keys(requiredFields).forEach((field) => {
     if (!payload[field]) {
-      throw new functions.https.HttpsError(
-        "invalid-argument",
-        requiredFields[field]
-      );
+      missingFields.push(field);
     }
   });
+
+  if (missingFields.length) {
+    throw new functions.https.HttpsError(
+      "invalid-argument",
+      HTTPSErrors.MissingParameter,
+      { missingFields }
+    );
+  }
 };
 
 /**
@@ -54,11 +51,7 @@ export const sendEmail = functions
     async ({ organization, ...email }: SendMailPayload, { auth }) => {
       await checkUser(organization, auth);
 
-      checkRequiredFields(email, {
-        to: SendEmailErrors.NoRecipient,
-        subject: SendEmailErrors.NoSubject,
-        html: SendEmailErrors.NoMsgBody,
-      });
+      checkRequiredFields(email, ["to", "subject", "html"]);
 
       // add email to firestore, firing data trigger
       await admin
@@ -78,6 +71,9 @@ export const sendSMS = functions
   .region("europe-west6")
   .https.onCall(async ({ organization, ...sms }: SendSMSPayload, { auth }) => {
     await checkUser(organization, auth);
+
+    // check payload
+    checkRequiredFields(sms, ["sender", "to"]);
 
     // get SMS template data
     const orgData =
@@ -208,7 +204,7 @@ const runWithTimeout = async <T extends any>(
 ): Promise<T> => {
   // set a timeout boundry for a function
   const timeoutBoundary = setTimeout(() => {
-    throw new functions.https.HttpsError("aborted", HTTPErrors.TimedOut);
+    throw new functions.https.HttpsError("aborted", HTTPSErrors.TimedOut);
   }, timeout);
 
   const res = await fn();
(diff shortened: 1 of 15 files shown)


silviot:
> Fixes #368

That's a strange loop indeed ;-)

eisbuk-admin/functions/src/https.ts
```diff
@@ -18,83 +18,71 @@ import {
...
+) => {
+  if (!payload) {
+    throw new functions.https.HttpsError(
+      "invalid-argument",
+      HTTPErrors.NoPayload
+    );
+  }
+
+  Object.keys(requiredFields).forEach((field) => {
+    if (!payload[field]) {
```
silviot, line 41: When more than one field is missing we'll be warned only about one of them. It would be nice to be told: _You're missing field A, B and C_ in one go instead.

eisbuk-admin/functions/src/https.ts
```diff
@@ -18,83 +18,71 @@ import {
...
+ * Stores email data to `emailQueue` collection, triggering firestore-send-email extension.
+ */
+export const sendEmail = functions
+  .region("europe-west6")
+  .https.onCall(
+    async ({ organization, ...email }: SendMailPayload, { auth }) => {
+      await checkUser(organization, auth);
+
+      checkRequiredFields(email, {
+        to: SendEmailErrors.NoRecipient,
```
silviot, line 60: I propose we use a single `MissingParameter` exception, and when we instantiate and throw it, we specify the list of fields that were missing.

silviot:
@ikusteu I pushed a new commit that uses [expect](https://www.oreilly.com/library/view/exploring-expect/9781565920903/) to answer `N` to the question `Would you like to proceed with deletion? Selecting no will continue the rest of the deployments.` that shows up when unexpected functions are found in production.

After deploying and trying to send a link via SMS I got:
```json
{
  "error": {
    "details": {
      "missingFields": [
        "0",
        "1"
      ]
    },
    "message": "One or more required parameters are missing from the payloady",
    "status": "INVALID_ARGUMENT"
  }
}
```

eisbuk-admin/functions/src/utils.ts
```diff
@@ -186,3 +186,75 @@ export const sendRequest = <
...
+  const missingFields: string[] = [];
+
+  Object.keys(requiredFields).forEach((field) => {
+    if (!payload[field]) {
+      missingFields.push(field);
+    }
+  });
+
+  if (missingFields.length) {
+    throw new EisbukHttpsError("invalid-argument", errorMessage, {
```
silviot, line 256: I believe this should be `new EisbukHttpsError("invalid-argument", errorMessage, missingFields)`.

eisbuk-admin/functions/src/utils.ts
```diff
@@ -186,3 +186,75 @@ export const sendRequest = <
...
+ * Throws error if one or more of the fields aren't present
+ * @param payload record to validate
+ * @param requiredFields array of field names (keys)
+ * @param errorMessage (optional) customer error message on failure, falls back to
+ * `HTTPSErrors.MissingParameter`
+ */
+export const checkRequiredFields = (
+  payload: Record<string, any> | undefined | null,
+  requiredFields: string[],
+  errorMessage: string = HTTPSErrors.MissingParameter
```
silviot, line 241: This parameter is currently never used. 
It would be nice to see the missing fields enumerated in the error message.
I think not having the message as a parameter would make for a better dev experience. The function should build an error message that states what fields were missing.

eisbuk-admin/functions/src/utils.ts
ikusteu: I dispute this...kinda. `details` is fixed to `functions.https.HttpsError` creator (which is the actual type of `EisbukHttpsError`), so doing `{ missingFields: .... }` provides more detail in why the error was thrown (otherwise we would get `details: ["field-1", "field-2"]`) which, in my opinion is not as descriptive.

The problem stands that I should fix the way the missing fields are reported: should be field names, not indexes (ofc) which I think has to do with passing of missing fields into `missingFields` array, rather than error construction.

eisbuk-admin/functions/src/utils.ts
ikusteu: I left it like this in case we want to pass a different error to the same checking logic. One possible use case I had in mind was when checking template data (which should be stored in firestore) so we would still pass template data object and required fields into the `checkRequiredFields`, but would return a different error: not saying that some parameters are missing, but stating that some tamplate data is missing from admin setup (not from request payload)

silviot: I think discussing the merits of the extra parameters for future usage would drive the conversation away from the main point. In my opinion we should have that discussion when the extra parameter comes into use. I think it would be better to introduce it when we start using it, but I won't put my foot down about this.

My point is that the `message` an exception carries should hold all the most relevant information regarding what error occurs. I believe that's very important. In some contexts only the message will be available. I will put my foot down about this.

As things stand, as a (dev) user of the `checkRequiredFields` function, I can't get it to output a message that summarizes what went wrong. `checkRequiredFields`, you had one job!

Again: I'm slightly against coding functionalities that will be used later (I fear that, lacking an actual use case, some problems might be missed), but what bothers me the most is that we don't get the missing field names in the message, and I believe we should.

If we don't include the missing fields in the error message we should have a strong reason not to do so. I fail to see such reason.

eisbuk-admin/functions/src/utils.ts
silviot: Agreed! I misread the code and commented thinking it was `{... missingFields}` but it's actually `{missingFields}`.

silviot (changes requested):
I tried to send an SMS using the preview channel (after deploying cloud functions from this branch to production).
I get this error:
```json
{
  "error": {
    "details": "The function must be called while authenticated with a user that is an admin of the given organization.",
    "message": "Unathorized",
    "status": "PERMISSION_DENIED"
  }
}
```
But I'm definitely logged in as an admin.

silviot:
After the failure above I compared the output of Chrome's `Copy as cURL` for the two requests. They were almost identical.
I swapped in the authorization header and the "bad one" started working.
I believe it's an edge case and we shouldn't worry about it. But I'm pissed I can't try out cloud functions invocation from dev channels!

silviot (approved):

PR #377, opened 2022-02-18
Fix/orphaned bookings

### LINKED ISSUE #364: Make customers and bookings document more connected through dataTrigger
Currently the bookings entry for a customer will get created/updated, but not deleted when a customer is deleted. This caused some "orphaned" bookings to exist in production/staging data. Therefore,  we should:
- [ ] Delete the bookings entry when customer gets deleted: full deletion shouldn't happen in production (as "deletion" actually adds `deleted: true` property to a customer), but we should at least pass `deleted: true` to bookings entry as well
- [ ] Create a migration function to delete all "orphaned" bookings

ikusteu:
Fixes #364


ikusteu:
There's a skipped (flaky) test...however, the functionality works and I've skipped the test not to waste any more time on testing a fix which should have been a very quick one (and test taking more time than development)

eisbuk-admin/functions/src/migrations.ts
```diff
@@ -383,3 +383,36 @@ export const addIdsToCustomers = functions
...
+    const allCustomers = await orgRef
+      .collection(OrgSubCollection.Customers)
+      .get();
+    const customerIds = allCustomers.docs.map(({ id }) => id);
+    const bookingsRef = orgRef.collection(OrgSubCollection.Bookings);
+    const allBookings = await bookingsRef.get();
+
+    const toDelete: Promise<any>[] = [];
+    allBookings.forEach((doc) => {
+      const docRef = doc.ref;
```
silviot, line 407: Very nitpicky, but I'll still mention this. No need to change (it can go in as is).

This line is here so that `docRef` can be used as a name, maybe because it's more descriptive, or shorter than `doc.ref`. I believe this line is not needed, and we can use `doc.ref` later in this function.

eisbuk-admin/src/pages/debug/index.tsx
```diff
@@ -61,7 +61,7 @@ const DebugPage: React.FC = () => {
       </Box>
       <Box my={4} color="secondary.main">
         <Button
-          onClick={invokeFunction(CloudFunction.CreateTestSlots)}
+          onClick={() => invokeFunction(CloudFunction.CreateTestSlots)()}
```
silviot, line 64: I don't get the reason for this change.

`invokeFunction` might be poorly named: it makes you think that when you invoke `invokeFunction` a cloud function will be invoked, as a side effect. But instead a function is returned that can be used to invoke the cloud function.
This naming can be improved, but in general code like

```javascript
const something = () => somefunction()();
```

should be written like this instead:
```javascript
const something = somefunction();
```

silviot (changes requested):
I think the `invokeFunction` change should be rolled back (and possibly `invokeFunction` could get a new name). I also think the `docRef` → `doc.ref` change would be good.

eisbuk-admin/functions/src/migrations.ts
ikusteu: +1

eisbuk-admin/src/pages/debug/index.tsx
ikusteu: I don't have an idea why, either. However, when invoking the "old" way, we get the console.error "max call stack size increased" and function (cloud function) doesn't get called. I believe it might have something to do with React's rendering cycle and maybe `useCallback` could fix this issue, but it's more straighforward this way...

ikusteu: Try the existing `master` and I believe you will be able to reproduce (max call stack size)

silviot: Thanks for the explanation and fix. I think I noticed the problem but didn't get to fix it.
I'm glad you took care of it!

PR #394, opened 2022-02-27
Cleanup/archive old migrations

### LINKED ISSUE #389: Remove old migrations and make sure https cloud functions require admin status
We have a bunch of migrations that, once run, are not needed anymore.
We should remove the buttons from the `/debug` page, and possibly remove their code. Or, since the code can provide useful hints when working on a new migration, we could move them to a different file that is not imported, and hence will not be deployed.

Let's also add `checkUser` to `unifyOffIceLabels` [that is currently missing it](https://github.com/eisbuk/EisBuk/blob/master/eisbuk-admin/functions/src/migrations.ts#L428).

ikusteu:
Fixes #389


ikusteu:
This is ready to go in, but to make it complete (not draft) we should:
- [ ] Run new (one time) migrations against production data
- [ ] Remove off-ice fallbacks from the app (as they'll become obsolete after migrations are ran)
- [ ] Archive the new migrations as well

silviot (approved):

PR #393, opened 2022-02-26
Cleanup/remove cra bump deps

ikusteu:
Fixes #392

PR #404, opened 2022-03-02
Remove react-scripts

silviot:
As a follow up to #393 this PR removes the `react-scripts` dependency.
I'm not sure why `tsc` now complains about the `svg`s. @ikusteu Do you happen to know why?


ikusteu:
Not really...but I have some ideas and can check it out.

ikusteu (approved):

PR #408, opened 2022-03-11
Paginate off-ice by month

### LINKED ISSUE #401: Off-ice booking view should be paginated by month instead of by week

fadwamahmoud:
Fixes #401 
Current behaviour (might be a bug): 
Both ice and off-ice views now show the same dates.
For `ice` the dates with both `ice` and `off-ice` slots are shown but only dates with `ice` slots actually display slots and vice versa for `off-ice`.


silviot:
The ordering of dates is off:
Screenshot from 2022-03-13 13-07-06

fadwamahmoud:
> The ordering of dates is off: slots are sorted by time of day, regardless of their date.
> 
> Screenshot from 2022-03-13 13-07-06

I didn't know whether to sort them or not because they weren't being sorted originally on off-ice view, they were just displayed as is from the selector. However, I added a function that sorts them by date.

silviot (approved):

eisbuk-admin/src/pages/customer_area/CustomerSlots/CustomerSlots.tsx
```diff
@@ -59,22 +59,20 @@ const CustomerSlots: React.FC<Props> = ({
   const slotDates = Object.keys(slots);
 
   // if `view=book_ice` should order slot days mondays first and so on
-  // if `view=book_off_ice` display week in standard order
+  // if `view=book_off_ice` display in standard order
   const orderedDates =
-    view === CustomerRoute.BookIce ? orderByWeekDay(slotDates) : slotDates;
+    view === CustomerRoute.BookIce
+      ? orderByWeekDay(slotDates)
+      : orderByDate(slotDates);
```
silviot, line 66: ISO dates sort lexycographically, so converting them to `DateTime` using the `fromISO` function is not needed (it's currently done in the `orderByDate` function). `slotDates.slice().sort()` should be enough

fadwamahmoud: Why `slice()` though? @silviot

eisbuk-admin/cypress/integration/booking_spec.ts
```diff
@@ -156,7 +156,10 @@ describe("Booking flow", () => {
       );
     });
 
-    it("disables only the locked bookings if locked/non-locked month overlap shown", () => {
+    /**
+     * @TODO make sure this test is necessary
+     */
+    it("enable booking in unlocked month and disable booking in locked month ", () => {
```
ikusteu, line 162: You're right to ask yourself that...it isn't necessary

PR #409, opened 2022-03-12
Feature/auth UI

ikusteu:
Our own auth UI dialog. The idea is to implement a phone login, compatible with firebase v9. And the UI and the logic is there and e2e tests everything works. In production, however, the reCaptcha is bugged.


eisbuk-admin/cypress/plugins/firebasePlugin/getSigninLink.ts
```diff
@@ -0,0 +1,69 @@
...
+    // search from end to front (to find the latest code)
+    .reverse()
+    .find(
+      ({ email, requestType }) =>
+        email === requestEmail && requestType === "EMAIL_SIGNIN"
+    ).oobCode;
+
+  // normally, the signin link would go to auth emulator (port 9099) and redirected back to the dev server (port 3000)
+  // cypress (browser actually), however, doesn't allow switching origin in the test, so we're manually constructing a login link
+  // going directoy to dev server
```
silviot, line 41: Typo

eisbuk-admin/cypress/support/index.ts
```diff
@@ -30,3 +30,8 @@ beforeEach(() => {
   );
   cy.stub(reactI18next, "useTranslation").callsFake(() => i18n.t);
 });
+
+afterEach(() => {
+  // auth from previous tests shouldn't affect the following tests
+  cy.signOut();
```
silviot, line 36: No objections to having this, sine explicit is better than implicit, but I'd like to validate my understanding of how Cypress works. Isn't the browser already reset to a clean state between tests, making this not strictly necessary?

silviot (approved):

PR #410, opened 2022-03-13
Enhancement/email link signin

### LINKED ISSUE #381: Switch to email link sign in for emails
We should make sure that users can't create accounts for emails they do not control (as is the case currently). One way might be [switching to link sign in](https://firebase.google.com/docs/auth/web/email-link-auth).

ikusteu:
Fixes #381 
Depends on #409 but can be merged on it's own (contains all 409 commits)


eisbuk-admin/src/hooks/useAuthFlow.tsx
```diff
@@ -77,6 +77,8 @@ const useAuthFlow: UseAuthFlow = (fieldErrorMap) => {
           // if error message not defined for a particular error code, fall back to unknown error
           t(AuthErrorMessage[error.code] || AuthErrorMessage.UNKNOWN);
 
+        console.log("Error code", error.code);
```
silviot, line 80: Is this supposed to still be here? Maybe it was supposed to be removed.

ikusteu: Yeah...will remove that

silviot (approved):

PR #418, opened 2022-03-21
Enhancement/date switcher

### LINKED ISSUE #388: Date switcher improvements
The date switcher from #384 should show days with data in them differently (a circle around them, a different colour or something similar).
For each view this might mean something different:
* slots view should show days with no slots/days with slots
* attendance view should distinguish three cases: no slots/ some slots but no bookings/ some slots that have some bookings
* athlete bookings should be similar, with three cases: no slots/slots available but none booked/some slots booked

fadwamahmoud:
Fixes #388


eisbuk-admin/src/components/atoms/DateSwitcher/DateSwitcher.stories.tsx
```diff
@@ -20,6 +20,7 @@ export const Default = (): JSX.Element => (
       open={true}
       anchorEl={null}
       onClose={() => {}}
+      jump={"week"}
```
ikusteu, line 23: A small thing: this could just be `jump="week"` - if you have a simple string, there's no need for curly braces...there was an eslint rule for this, but we obviously don't implement that :thinking:

eisbuk-admin/src/store/selectors/calendar.ts
```diff
@@ -0,0 +1,49 @@
+import { LocalStore } from "@/types/store";
+
+interface calendarSlot {
```
ikusteu, line 3: Types should be PascalCase, (i.e. `CalendarSlot`)...this is a convention, but it's also there to distinguish the often used same naming of function and function interface (the example below is a very simplified usage, and, in that case, I would declare the function interface inline, but this format is used for more complex functions to visually separate type interface and function implementation for better readability)
```typescript
interface GetTwo {
  (): number
}

const getTwo: GetTwo () => 2
```

eisbuk-admin/src/store/selectors/calendar.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export const getCalendarData = (state: LocalStore): calendarSlot => {
+  const {
+    firestore: {
+      data: { slotsByDay: allSlotsInStore, attendance = {}, bookedSlots = {} },
+    },
+  } = state;
+
+  const slotsByDay = allSlotsInStore ?? {};
```
ikusteu, line 14: `const slotsByDay = allSlotsInStore ?? {}` is the same as declaring
```
const { firestore: {
    data: { slotsByDay = {} }
  }
} = state
```
in both cases, the slotsByDay will fall back to empty `{}` if `undefined | null`

ikusteu: Scratch that...it will only work if undefined, not if `null`

eisbuk-admin/src/store/selectors/calendar.ts
```diff
@@ -0,0 +1,49 @@
...
+      ? Object.values(attendance).reduce((acc, _) => {
+          const bookedIntervals = Object.values(_.attendances).filter(
+            (customerAttendance) => customerAttendance.bookedInterval
+          );
+          return bookedIntervals.length ? { ...acc, [_.date]: "booked" } : acc;
+        }, {})
+      : Object.values(bookedSlots).reduce(
+          (acc, _) => ({ ...acc, [_.date]: "booked" }),
+          {}
+        );
```
ikusteu, line 36: `_` should only be used as: 
- a write-only variable - a variable we don't plan on using but need to declare for syntax purposes, i.e.
```typescript
interface AFunction {
  (a: string, b: string, c: string): any
}
// a function which needs do declare the `b` to get to `c`
const returnAPlusC: AFunction = (a, _, c) => a + c
// a function which doesn't need `c` nor `b` so both are omitted (no need for write-only variable)
const idA: AFunction = (a) => a
```

eisbuk-admin/src/store/selectors/calendar.ts
```diff
@@ -0,0 +1,49 @@
...
+  };
+
+  const datesWithSlots: calendarSlot = Object.keys(flattenedSlotsByDay).reduce(
+    (acc, _) => {
+      return {
+        ...acc,
+        [_]: "slots",
+      };
+    },
```
ikusteu, line 45: In continuation of the `_` remark, here is makes perfect sense to name id `dateISO` and have the code be more expressive. I other cases, where it doesn't make much sense semantically, you can always use `curr` as in `(acc, curr) => ...`

eisbuk-admin/src/components/atoms/DateSwitcher/__tests__/DateSwitcher.test.tsx
```diff
@@ -45,5 +47,43 @@ describe("Date Switcher", () => {
...
+            anchorEl={null}
+            onClose={() => {}}
+            jump={"week"}
+          />
+        </LocalizationProvider>
+      );
+
+      const nexDayDateTime = testDateDateTime.plus({ days: 1 });
+
+      const dayToClick = screen.getByLabelText(nexDayDateTime.toFormat("DD"));
```
ikusteu, line 65: This could be one variable

eisbuk-admin/src/components/atoms/DateSwitcher/__tests__/DateSwitcher.test.tsx
```diff
@@ -14,6 +14,7 @@ import LocalizationProvider from "@mui/lab/LocalizationProvider";
 import { testDateLuxon } from "@/__testData__/date";
 import { changeCalendarDate } from "@/store/actions/appActions";
 import { DateTime } from "luxon";
+import { __PickedDay__ } from "@/__testData__/testIds";
```
ikusteu, line 17: to keep consistent casing, this should be `__pickedDay__`

eisbuk-admin/src/__testData__/testIds.ts
```diff
@@ -29,6 +29,10 @@ export const __currentDateId__ = "current-date";
 
 // #region DateSwitcher
 export const __calendarMenuId__ = "calendar-menu";
+export const __PickedDay__ = "picked-day";
+export const __DayWithSlots__ = "day-with-slots";
+export const __DayWithBookedSlots__ = "day-with-booked-slots";
```
ikusteu, line 34: To keep consistent casing, these should be cased `__camelCaseWithUnderscore__`

eisbuk-admin/cypress/integration/calendar_spec.ts
```diff
@@ -0,0 +1,66 @@
...
+  __DayWithBookedSlots__,
+  __DayWithSlots__,
+} from "@/__testData__/testIds";
+
+describe("Date Switcher", () => {
+  describe("Badges", () => {
+    beforeEach(() => {
+      // our test data starts with this date so we're using it as reference point
+      const testDate = "2022-01-01";
+      const testDateLuxon = DateTime.fromISO(testDate);
```
ikusteu, line 15: No need for two variables

eisbuk-admin/cypress/integration/calendar_spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+      cy.setClock(testDateLuxon.toMillis());
+      cy.initAdminApp().then((organization) =>
+        cy.updateFirestore(organization, [
+          "bookings.json",
+          "slots.json",
+          "customers.json",
+          "attendance.json",
```
ikusteu, line 23: Attendance is probably unnecessary as it's created/updated through slot creation data trigger. That way you reduce possible inconsistencies in test data and at the same time test as close to production setting as it gets

ikusteu (changes requested):
I wrote some small (papercut) comments, but I think we should go through a pair-programming session to have a new take on `getCalendarData` (and test data that comes with it)

eisbuk-admin/cypress/integration/calendar_spec.ts
ikusteu: But we can go over this in detail in PP session

ikusteu (approved):

PR #423, opened 2022-03-23
Add inputMode field to phone and PIN

fadwamahmoud:


silviot (approved):

ikusteu (approved):
Awesome...I especially like the `copy from messages` option (idk if that's an iPhone specivif thing, but thanks to this in type of input it just works)

PR #426, opened 2022-03-27
Cleanup/tidy monorepo

### LINKED ISSUE #313: Investigate dead code elimination options
Since migrating to ESBuild from CRA, the production produces the following error:
```
Uncaught Error: React is running in production mode, but dead code elimination has not been applied. Read how to correctly configure React for production: https://reactjs.org/link/perf-use-production-build
    at <anonymous>:139:17
```

Should see and add a solution with ESBuild (if it's just a config thing), or check the official solutions (post processing of the bundle), whichever is easier to implement.

### LINKED ISSUE #343: Fix sourcemaps for build/serve
We're using minified build for both `build` and `serve` commands (as `serve` was breaking without minification). However, it would be nice to have sourcemaps point to actual stack trace positions in code. This was working temporarily, but now it's buggy again. Should check and fix that.

ikusteu:
A work on tidying up the monorepo and setting groundwork for additional features (vite, pnpm, rushjs). This is the sensible amount of work ready to go in while I work on additional features. Current updates include:
- Create `scaffolding/` folder containing base config for TypeScript and ES-Lint (every other package extends those base configs with minimal additional setup
- Lift `functions/` package to root of the repo and thus make all of the packages siblings
- Use Vite for bundling/serving of `eisbuk-admin` and `eisbuk-shared` (ES-Build is still used for `functions` as Vite would be unnecessarily added layer of abstraction and complexity)

Thanks to Vite, this PR also:
Fixes #343 
Fixes #313 

_Next steps require additional work, and while I'm working on it, there are other features that are more of a priority. I will update this PR with any solved TODOs until it gets merged. When the PR is merged, the unsolved TODOs will become separate tickets/PRs._

**TODO:**
- [ ] Add storybook-builder-vite - I've already tried this, but it throws weird errors. I think it would definitely pay to utilize Vite's HMR, but this work is delayed until higher priority tasks are solved (and current Storybook setup works alright)
- [ ] Add pnpm - I've tried this, but wasn't able to make it work out of the box, so the situation is similar as with the aforementioned Storybook update
- [ ] Lift `cypress` to the root of the repo and make it a separate package
- [ ] Make all of the scripts callable from the root of the repo (i.e. `client:start`, `client:eisbuk`, `enulators:start`, etc...)
- [ ] Lift coverage logic to root of the repo and include all relevant sources

Some features from the start of the task which I feel is better to omit:
- Vite Jest - I don't really see the utility of Vite (fast HMR) in unit tests, only the ESBuild transpiler for speed, but that can be solved differently


.github/workflows/build.yml
```diff
@@ -27,7 +27,9 @@ jobs:
         run: |
           yarn --cwd eisbuk-admin build | tee build.output
           echo "BUILD_STATS<<EOF" >> $GITHUB_ENV
-          cat build.output|grep File\ sizes -A5 >> $GITHUB_ENV
+          # Select all lines starting with 'dist/' from build.output (those are the lines reporting the build file sizes)
+          # And store for use in later steps
+          cat build.output|awk '/^dist\// {print}' >> $GITHUB_ENV
```
silviot, line 32: `awk '/^dist\// {print}'` means: print any line startng with `dist/`, and print no other line.
I think it's more readable with `grep ^dist/`.

silviot (approved):

PR #443, opened 2022-04-12
Feature/add to calendar

### LINKED ISSUE #430: Add to calendar
Athletes should be able (after booking) to add their bookings to their calendar (iCal or Google Calendar).
This can happen either with downloadable iCal files or using APIs. The former is easier to set up, but won't allow automatic updates.

We should support Apple (iCal) and Google (Google Calendar) to support both Android and iOS users.

fadwamahmoud:
Fixes #430


packages/client/src/store/actions/organizationOperations.ts
```diff
@@ -13,17 +13,26 @@ import { enqueueNotification, showErrSnackbar } from "./appActions";
...
       const db = getFirestore();
       const docRef = doc(db, getOrganizationCollPath());
+      const publicOrgInfoDocRef = doc(db, getPublicOrganizationInfoCollPath());
+
       await setDoc(docRef, orgData, { merge: true });
+      await setDoc(
+        publicOrgInfoDocRef,
+        { displayName, location, emailFrom },
+        { merge: true }
+      );
```
ikusteu, line 35: Right...so I really wouldn't do it this way, but would rather store all of the data in organization document (`organizations/{orgName}`) and use a data trigger to (selectively) copy only the data that should be publicly available, over to `publicOrgInfo/{orgName}` (or which ever the path). 
The reason for this is that it simplifies the client's role in updating the organization data (and thus separates the concerns in somewhat of a right way) and enables us to reuse the data from org in both restricted mode (admin-only data) as well as public mode (available to non-registered users). An example of this would be a `from` email: used in admin-mode to send emails as well as in non-admin mode to display "Not registered, contact this email..." message.

To summarize, the steps I would take:
- [ ] Add all data to org info (organization document)
- [ ] Create a data trigger to selectively copy the data to publicly available doc 
- [ ] Reflect this in `firestore.rules`:
- anybody can read - `read: if true;`
- nobody can write - `write: if false; // or just the absence of write rule should do it` (data triggers and cloud functions aren't blocked by this rule)

packages/client/src/store/actions/organizationOperations.ts
```diff
@@ -37,6 +46,7 @@ export const updateOrganization =
       );
       setSubmitting(false);
     } catch (error) {
+      console.log({ error });
```
ikusteu, line 49: If we're doing this, it should be `console.error`. It's always disputable whether or not we should log to console in production, but this way the sentry will pick it up as an error (I think), so we get an added benefit there.

packages/client/src/pages/admin_preferences/index.tsx
```diff
@@ -98,7 +98,7 @@ const OrganizationSettings: React.FC<Props> = () => {
               <Divider />
               <Form className={classes.form}>
                 <FormControl component="fieldset">
-                  <FormSection content={nameField} />
+                  <FormSection content={publicInfoFields} name="Public Info" />
```
ikusteu, line 101: I wouldn't name this "Public info" but rather just "General". Since admin using our app doesn't need to care not know which data will be available to whom (in this sense). It's up to the app to make sure that appropriate data is available to appropriate parties, without bothering the user with the implementation. 
Additionally there might be some "General" data which won't be available to the public. At which point, the distinction would be made on the implementation level (as I've described with the data trigger logic).

Also, this should be translateble, so it's best to provide the translation logic right away and simply add the italian translation later.

packages/client/src/store/selectors/bookings.ts
```diff
@@ -43,6 +43,22 @@ export const getBookedSlots = (
...
+  (month: number) =>
+  (state: LocalStore): Record<string, CustomerBookingEntry> =>
+    Object.entries(state.firestore.data?.bookedSlots || {}).reduce(
+      (acc, curr) => {
+        return DateTime.fromISO(curr[1].date).month === month
+          ? { ...acc, [curr[0]]: curr[1] }
+          : acc;
+      },
+      {}
+    );
```
ikusteu, line 61: Loving the `Object.entries` usage. I wasn't aware of it at all and it simplifies things a lot. However, to produce more semantic readability, I would destructure the `[key, value]` entires, (rather then using `curr[0]` and `curr[1]`):

```typescript
// Instead of writing
Object.entries(state.firestore.data?.bookedSlots || {}).reduce(
  (acc, curr) => {
    return DateTime.fromISO(curr[1].date).month === month
      ? { ...acc, [curr[0]]: curr[1] }
      : acc;
  },
  {}
);

// I would write
Object.entries(state.firestore.data?.bookedSlots || {}).reduce(
  (acc, [slotId, bookingEntry]) => {
    return DateTime.fromISO(bookingEntry.date).month === month
      ? { ...acc, [slotId]: bookingEntry }
      : acc;
  },
  {}
);
```

ikusteu: Also, notice the omitted `{ return /* ... */ }`. It's unnecessary.

packages/client/src/types/store.ts
```diff
@@ -164,6 +164,10 @@ export interface FirestoreData {
   };
   [OrgSubCollection.SlotsByDay]: { [monthStr: string]: SlotsByDay } | null;
   [OrgSubCollection.Attendance]: { [slotId: string]: SlotAttendnace };
+  [Collection.PublicOrgInfo]: {
+    [organization: string]: Pick<OrganizationData, "displayName"> &
+      Pick<OrganizationData, "location">;
+  };
```
ikusteu, line 170: This goes hand in hand with my suggestion: The `publicOrgInfo` is a subset of `OrganizationData`. Also, what I've lately discovered, the long `Pick` subset can be written (shortly) like this: `Pick<OrganizationData, "displayName" | "location">`

packages/e2e/integration/calendar_spec.ts
```diff
@@ -85,3 +89,31 @@ describe("Date Switcher", () => {
     });
   });
 });
+describe("Add To Calendar", () => {
+  describe("ics file", () => {
+    const downloadsFolder = Cypress.config("downloadsFolder");
+    beforeEach(() => {
+      cy.setClock(testDateLuxon.toMillis());
```
ikusteu, line 96: No need for double describe if there's only one test.

packages/e2e/integration/calendar_spec.ts
```diff
@@ -22,7 +26,7 @@ const openCalendar = (
   jump: DateTimeUnit = "week"
 ) => cy.contains(i18n.t(createDateTitle(date, jump)) as string).click();
 
-describe("Date Switcher", () => {
+xdescribe("Date Switcher", () => {
```
ikusteu, line 29: Why the `xdescribe`?

packages/e2e/plugins/index.ts
```diff
@@ -30,6 +31,22 @@ module.exports = (on: Cypress.PluginEvents, config: Cypress.PluginConfig) => {
...
+        rmdir(folderName, { maxRetries: 10, recursive: true }, (err) => {
+          if (err && err.code !== "ENOENT") {
+            console.error(err);
+            return reject(err);
+          }
+          resolve(null);
+        });
+      });
+    },
+  });
```
ikusteu, line 49: Let's keep the task handlers in their own plugin(s). For this use case, I suggest you make a `fsPlugin` and follow the `firebasePlugin` structure. 
Please note: This is not a generally accepted convention for cypress plugins, but it made a lot of sense to me when writing the first one (`firebasePlugin`) and I believe we should keep things consistent, as well as modular (not writing plugins directly in `plugins/index.ts`)

packages/e2e/integration/calendar_spec.ts
```diff
@@ -85,3 +89,31 @@ describe("Date Switcher", () => {
     });
   });
 });
+describe("Add To Calendar", () => {
+  describe("ics file", () => {
```
ikusteu, line 93: No need for double describe and a `beforeEach` block if there's only one test

packages/shared/src/types/firestore.ts
```diff
@@ -76,6 +76,10 @@ export interface OrganizationData {
    * for a given organization
    */
   existingSecrets?: string[];
+  /**
+   * Organization Location
+   */
```
ikusteu, line 81: Maybe provide some more context to "Organization Location"...this might be used as doc (or when writing the docs) so it should be clear what it is.

packages/translations/src/translations.ts
```diff
@@ -203,7 +205,7 @@ export enum NotificationMessage {
 
   EmailSent = "Notification.EmailSent",
   SMSSent = "Notification.SMSSent",
-  BookingDateExtended = "Notification.BookingdateExtended",
+  BookingDateExtended = "Notification.BookingDateExtended",
```
ikusteu, line 208: Touche

packages/e2e/__testData__/bookedSlots.json
```diff
@@ -0,0 +1,10 @@
+{
+  "slot-1": {
+    "date": "2022-01-01",
+    "interval": "09:00-10:00"
+  },
+  "slot-2": {
+    "date": "2022-01-02",
+    "interval": "10:00-11:00"
+  }
+}
```
ikusteu, line 10: This doesn't seem to be used anywhere...and is of a wrong format (wouldn't work with `updateFirestore`) anyhow.

packages/client/src/pages/customer_area/CustomerSlots/CustomerSlots.tsx
```diff
@@ -141,9 +150,6 @@ const CustomerSlots: React.FC<Props> = ({
           </SlotsDayContainer>
         );
       })}
-      {countdownProps?.message === BookingCountdownMessage.SecondDeadline && (
-        <BookingsCountdown {...countdownProps} />
-      )}
```
ikusteu, line 146: Why delete this?

packages/client/src/components/atoms/BookingsCountdown/__tests__/AddToCalendar.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+jest.spyOn(reactRedux, "useDispatch").mockImplementation(() => mockDispatch);
+
+describe("Add To Calendar", () => {
+  describe("Smoke Test", () => {
+    test("should render component", () => {
+      renderWithRedux(<AddToCalendar />);
+      screen.getByTestId(__addToCalendarButtonId__);
+    });
+  });
+});
```
ikusteu, line 23: Unless you plan to extend this test, I would say the entire file is completely unnecessary.

packages/client/src/components/atoms/BookingsCountdown/AddToCalendar.tsx
```diff
@@ -0,0 +1,113 @@
+import React from "react";
```
ikusteu, line 1: This has nothing to do with `BookingsCountdown` and should be in it's own folder...same as the tests. So do the standard `index.ts`, `<ComponentName>.tsx`,  `__tests__/<ComponentName>.test.tsx` folder structure.

packages/client/src/components/atoms/BookingsCountdown/AddToCalendar.tsx
```diff
@@ -0,0 +1,113 @@
...
+        icalendar
+          .addEvent(new ICalendar(bookedSlotEvent))
+          .addProperty("UID", `${bookedSlot.date}${bookedSlot.interval}`);
+      }
+      /**
+       * Download calendar on last iteration
+       */
+      if (i === Object.keys(bookedSlots).length - 1) {
+        icalendar.download("Booked_Slots.ics");
+      }
```
ikusteu, line 67: Instead of counting `bookedSlots`' keys (and performing additional iteration, you can access the array you're already iterating over (`Object.values(bookedSlots)`), using a third `.forEach` param. i.e.
```typescript
 Object.values(bookedSlots).forEach((bookedSlot, i, originalArr) => {
      // ...rest of the code
      const arrLen = originalArr.length;
      // ...rest of the code
    });
```

ikusteu: This way, you're not constructing a new array (`Object.keys(bookedSlots)`) just to count it.

ikusteu: Also, is the last iteration check really necessary...you can just download the calendar after the `forEach` loop...`calendar` is still in scope.

ikusteu (changes requested):
I left some comments...most important one is the idea of date triggers, then the folder structure and organization. After that, all the rest: unnecessary files, code performance/complexity, coding style, etc.

packages/client/src/pages/customer_area/CustomerSlots/CustomerSlots.tsx
```diff
@@ -110,6 +115,10 @@ const CustomerSlots: React.FC<Props> = ({
   const content = (
     <>
       {countdownProps && <BookingsCountdown {...countdownProps} />}
```
fadwamahmoud, line 117: This is why I deleted the other BookingsCountdown. It displays a duplicate when there is a second deadline. @ikusteu

ikusteu: Yes, that (two notifications onsecond deadline) is the desired behaviour: https://github.com/eisbuk/EisBuk/pull/413

packages/client/src/pages/admin_preferences/index.tsx
```diff
@@ -53,7 +53,7 @@ const emailFields = [
     multiline: true,
   },
 ];
-const nameField = [{ name: "displayName" }];
+const GeneralFields = [{ name: "displayName" }, { name: "location" }];
```
ikusteu, line 56: Non-class variable names should be `camelCased`

packages/client/src/store/selectors/bookings.ts
```diff
@@ -43,6 +43,21 @@ export const getBookedSlots = (
...
+/**
+ * Get subscribed slots from state for a specific month
+ * @param state Local Redux Store
+ * @returns record of subscribed slots
+ */
+export const getBookedSlotsByMonth =
+  (month: number) =>
+  (state: LocalStore): Record<string, CustomerBookingEntry> =>
+    Object.entries(state.firestore.data?.bookedSlots || {}).reduce(
+      (acc, [key, value]) =>
```
ikusteu, line 55: I still believe that `slotId` and `bookingEntry` naming would be more appropriate and self documenting than `key`, `value`

packages/e2e/plugins/fsPlugin/index.ts
```diff
@@ -0,0 +1,22 @@
...
+    deleteFolder(folderName) {
+      console.log("deleting folder %s", folderName);
+
+      return new Promise((resolve, reject) => {
+        // eslint-disable-next-line consistent-return
+        rmdir(folderName, { maxRetries: 10, recursive: true }, (err) => {
+          if (err && err.code !== "ENOENT") {
+            console.error(err);
+            return reject(err);
+          }
```
ikusteu, line 14: This feels a bit wrong...What happens is that if (and only if) err.code is `ENOENT` you console it and `reject()`, which (and correct me if I'm wrong) is the promise equivalent of throwing errors. There are two points:
1. I feel it would be ok to not fail if the folder doesn't exist (and just `console.log` it, so it's not as silent)
2. If there's another error, I believe you would like it to fail (by throwing an exception) rather than failing silently an messing up the tests (for no apparent reason)

ikusteu: Btw. :+1: :+1: on using the separate plugin file

packages/e2e/integration/calendar_spec.ts
```diff
@@ -85,3 +89,28 @@ describe("Date Switcher", () => {
...
+    cy.initAdminApp().then((organization) =>
+      cy.updateFirestore(organization, [
+        "saul_with_extended_date.json",
+        "slots.json",
+      ])
+    );
+    cy.task("deleteFolder", downloadsFolder);
+    cy.visit([Routes.CustomerArea, saul.secretKey, "book_ice"].join("/"));
+    cy.getAttrWith("aria-label", "See past dates").click();
+    cy.getAttrWith("aria-label", "Book").first().click();
```
ikusteu, line 107: Could you please use the aria label enums (like Chris did at some point, slot form I think...), so that we have that covered when merging this in.

packages/shared/src/types/firestore.ts
ikusteu: Could you please provide more context on organization location, i.e. the location of ice skating rings, or the address of the ice skating club, or something in the ball park.

packages/client/src/components/atoms/AddToCalendar/AddToCalendar.tsx
```diff
@@ -0,0 +1,113 @@
...
+        icalendar
+          .addEvent(new ICalendar(bookedSlotEvent))
+          .addProperty("UID", `${bookedSlot.date}${bookedSlot.interval}`);
+      }
+      /**
+       * Download calendar on last iteration
+       */
+      if (i === originalArr.length - 1) {
+        icalendar.download("Booked_Slots.ics");
+      }
```
ikusteu, line 67: Can we just download the calendar after the loop is finished (that way there's no need to comment/explain anything there)

packages/client/src/components/atoms/AddToCalendar/AddToCalendar.tsx
```diff
@@ -0,0 +1,113 @@
...
+       */
+      if (i === 0) {
+        icalendar = new ICalendar(bookedSlotEvent).addProperty(
+          "UID",
+          `${bookedSlot.date}${bookedSlot.interval}`
+        );
+      } else {
+        /**
+         * To add another event to the calendar it must be of type ICalendar
+         */
```
ikusteu, line 57: Please use `// Comment` comments for in code explanations, and `/** JSDoc here... */` comments for JSDoc. That is the convention and I believe, in this case, it also reduces the number of lines and makes code a tad more readable.

packages/client/src/components/atoms/AddToCalendar/__tests__/AddToCalendar.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+jest.spyOn(reactRedux, "useDispatch").mockImplementation(() => mockDispatch);
+
+describe("Add To Calendar", () => {
+  describe("Smoke Test", () => {
+    test("should render component", () => {
+      renderWithRedux(<AddToCalendar />);
+      screen.getByTestId(__addToCalendarButtonId__);
+    });
+  });
+});
```
ikusteu, line 23: This file is still unnecessary:
- as I understand the complete functionality is tested with cypress
- as much as I believe in unit tests (for TDD), even though the feature is later tested in e2e, here you're just testing that something exists in screen, and it's not even something visible, but rather a test id created and assigned only to test it being assigned

packages/client/src/__tests__/customerDataTriggers.test.ts
```diff
@@ -26,6 +26,12 @@ import { saul } from "@/__testData__/customers";
 
 const customersCollectionPath = `${Collection.Organizations}/${getOrganization()}/${OrgSubCollection.Customers}`;
 const bookingsCollectionPath = `${Collection.Organizations}/${getOrganization()}/${OrgSubCollection.Bookings}`;
+const publicOrgPath = `${Collection.PublicOrgInfo}/${getOrganization()}`;
+
+const publicOrgRef = getDocumentRef(
+  adminDb,
+  publicOrgPath
+);
```
ikusteu, line 34: This seems unrelated to the file. If it's something you've forgotten to delete (after adding a different organization for calendar testing I imagine), please delete it. If not (and this was the only way to make this work), leave a `@TODO` comment to fix this more cleanly in the future (I'll do it with tests chore ticket)

packages/client/src/__tests__/dataTriggers.test.ts
```diff
@@ -332,4 +333,42 @@ describe("Cloud functions -> Data triggers ->,", () => {
...
+
+        const publicOrgPath = `${Collection.PublicOrgInfo}/${orgName}`;
+        const orgPath = `${Collection.Organizations}/${orgName}`;
+
+        // check for non existence of publicOrgInfo before creating a new organization
+        const publicOrgRes = await waitForCondition({
+          documentPath: publicOrgPath,
+          condition: (data) => Boolean(!data),
+        });
+        expect(publicOrgRes).toBeUndefined();
```
ikusteu, line 351: Right...this check makes sense when you're writing, or debugging the test (during a TDD session), but if you've set up the tests correctly, shouldn't be necessary: The purpose of the test isn't to test whether or not the organization entry is empty.

packages/client/src/components/atoms/FormSection/FormSection.tsx
```diff
@@ -28,16 +28,17 @@ const FormSection: React.FC<Props> = ({ name, content }) => {
 
   return (
     <div>
-      {name && <h5 className={classes.sectionTitle}>{name}</h5>}
+      {name && (
+        <h5 className={classes.sectionTitle}>{t(OrganizationLabel[name])}</h5>
+      )}
```
ikusteu, line 33: Here you're losing the whole point of enums (autocomplete and restricting your choice to enums which actually exist). The name should just be a string and you should take care of the translation when before passing that string. When rendering the component and passing a `name` prop into it.

ikusteu: Organization label, down in the code is a different matter and, unfortunately, there's no clever way about it (also, it's an enum to follow the consistency of using enum translations). 
The only way to make this type safe would be to update the `FieldProps` as such:
```typescript
interface FieldProps {
  name: keyof OrganizationData;
  label: keyof typeof OrganizationLabel;
  multiline?: boolean;
}
```

ikusteu: So...this way, you wouldn't use any transformations on the name -> label, but would rather read both the name and the label from `FieldProps` and pass the to translation enum, well assured that the enum key, in fact, exists.

ikusteu: If this is unclear (as I'm really tired), hit me up tomorrow for further explanation :joy:

packages/firebase/firestore.rules
```diff
@@ -193,5 +193,10 @@ service cloud.firestore {
       }
       // #endregion attendance-checks
     }
+    // #region public-info-check
+      match /publicOrgInfo/{organization} {
+      allow read: if true;
+      }
+    // #endregion public-info-check
```
ikusteu, line 200: Please watch the indentation here (prettier doesn't work in `.rules` files), otherwise this will get really messy really quickly

packages/functions/src/dataTriggers.ts
```diff
@@ -311,3 +312,35 @@ export const registerCreatedOrgSecret = functions
...
+      .doc(organization);
+    const orgData = change.after.data() as OrganizationData;
+
+    if (isDelete) {
+      await publicOrgInfoDocRef.delete();
+      return;
+    } else if (
+      !orgData.displayName ||
+      !orgData.location ||
+      !orgData.emailFrom
```
ikusteu, line 337: This way only if all of the values are updated, they would be updated to public org...we don't want that:
- cleaner solution, `O(n)`, would be to check all three if they exist and add them to an update object:
```typescript
const updates = ["displayName", "location", "emailFrom"].reduce(
  (acc, curr) => orgData[curr] ? {...acc, [curr]: orgData[curr]} : acc
)
await publicDocRef.set(updates, {merge: true)
```
- a less clean, but slightly more efficient one, `O(1)` (avoiding the looping):
```typescript
await publicOrgInfoDocRef.set({
      displayName: orgData.displayName || "",
      location: orgData.location || "",
      emailFrom: orgData.emailFrom || "",
    })
```
Needless to say, since it's only three fields (`n = 3` in `O(n)` and this is JS and not C rocket science, I would definitely go with the former (cleaner)

ikusteu (changes requested):
Left some updated comments, hit me up if you need any clarification, and please dispute the points where I'm not sure I'm right.

packages/e2e/plugins/fsPlugin/index.ts
fadwamahmoud: but what happens here is that it only `rejects` and `console.error` if the error code is anything *but* `ENOENT`

packages/client/src/components/atoms/AddToCalendar/AddToCalendar.tsx
fadwamahmoud: it's getting instantiated inside the loop though so it couldn't be used outside @ikusteu

packages/client/src/components/atoms/AddToCalendar/__tests__/AddToCalendar.test.tsx
fadwamahmoud: In your previous comment you mentioned to leave it be if I was planning to add more tests, which I am. In my next PR `feature/updatable_calendar_events` I write another test.
I could delete the file from this PR if that makes more sense.

packages/e2e/plugins/fsPlugin/index.ts
ikusteu: Right…sry I was tired lol

packages/client/src/components/atoms/AddToCalendar/AddToCalendar.tsx
ikusteu: It’s instantiated in the loop, but is declared outside the loop, therefore, it’s accessible from outer scope of the loop. I don’t know why TypeScript complains about it…but you can set it to an empty object when you declare it and cast the type, like so: ‘let calendar = {} as ICalendar’

packages/client/src/components/atoms/AddToCalendar/AddToCalendar.tsx
```diff
@@ -0,0 +1,106 @@
...
+  bookedSlots?: LocalStore["firestore"]["data"]["bookedSlots"];
+}
+const AddToCalendar: React.FC<Props> = ({ bookedSlots = {} }) => {
+  const classes = useStyles();
+
+  const { displayName = "", location = "" } =
+    useSelector(getAboutOrganization) || {};
+  const { t } = useTranslation();
+
+  // is displayed when bookings are finalized aka deadline is null
```
ikusteu, line 29: I don't understand this...I understand it might be an explanation of the component, but then it should be in component JSDoc..

ikusteu: Or just removed to avoid confusion if it was left here by accident

ikusteu (approved):

PR #451, opened 2022-04-20
Feature/import export

delasous:
This PR intends to add the functionality for a developer/admin to import-export production data from firestore via a CLI. This will be useful for backups, bulk-edits etc.


common/git-hooks/pre-commit
```diff
@@ -1,11 +0,0 @@
...
-# Called by "git commit" with no arguments.  The hook should
-# exit with non-zero status after issuing an appropriate message if
-# it wants to stop the commit.
-
-# Invoke the "lint-staged" configured in <repoRoot>/.linststagedrc.json.
-# .ts(x), .js(x) files are passed to `eslint` with no warnings allowed, and on success passed to `prettier write`.
-# .json, .yml, .yaml files are passed directly to `prettier write`
-# The command itself is defined in common/config/rush/command-line.json
-# and uses the "lint-staged" autoinstaller.
-node common/scripts/install-run-rush.js lint-staged || exit $?
```
ikusteu, line 11: Errm...we actually want that file...it's an equivalent of husky's pre-commit functionality

common/git-hooks/commit-msg.sample
```diff
@@ -1,25 +0,0 @@
...
-# an appropriate message if it wants to stop the commit.  The hook is allowed to edit
-# the commit message file.
-
-# This example enforces that commit message should contain a minimum amount of
-# description text.
-if [ `cat $1 | wc -w` -lt 3 ]; then
-  echo ""
-  echo "Invalid commit message: The message must contain at least 3 words."
-	exit 1
-fi
```
ikusteu, line 25: Yeah...I feel this one is not necessary, but if `rush update` keeps generating it, you can just leave it there.

packages/firestore/package.json-e
```diff
@@ -0,0 +1,7 @@
+{
```
silviot, line 1: Was this committed by mistake? I don't understand its significance.

packages/backup-restore/src/config/configstore.ts
```diff
@@ -0,0 +1,27 @@
+import fs from "fs";
+import Configstore from "configstore";
+
+const pkg = JSON.parse(fs.readFileSync("./package.json", "utf8"));
```
silviot, line 4: This means that, when the script is invoked, it should look up a file called `package.json` in the current directory.
This is not what we want.

packages/backup-restore/src/config/configstore.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export const defaultConfig: Config = {
+  projects: [],
+  activeProject: null,
+  useEmulators: "false",
+  emulatorHost: "localhost:8080",
+};
+
+export const configstore = new Configstore(pkg.name, defaultConfig);
```
silviot, line 27: There might be a way to get the package name injected here at compilation time, but it probably makes sense to just hardcode it here, replacing `pkg.name` with `"@eisbuk/backup-restore"`.

packages/backup-restore/src/commands/projects.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// TODO: following two functions could be fragmented and tested better
+/**
+ * Add firebase project credentials to XDG-specific AppData location
+ */
+export async function addProjectCredentials(filePath: string): Promise<void> {
+  // Resolve path to file and check it exists
+  const currentDir = process.cwd();
+  const resolvedPath = path.resolve(currentDir, filePath);
```
silviot, line 26: Non need to pass `currentDir` here: it's already the default. From `path.resolve` docs:
>  If after using all {from} paths still no absolute path is found, the current working directory is used as well.

packages/firestore/package.json-e
delasous: These are mistakes. I have no idea where they came from. I have removed them

packages/backup-restore/src/commands/projects.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  try {
+    const fileExists = await exists(resolvedPath);
```
silviot, line 29: The docs for [`fs.access`](https://nodejs.org/api/fs.html#fsaccesspath-mode-callback) (used in the `exists` function) say:

>Do not use fs.access() to check for the accessibility of a file before calling fs.open(), fs.readFile(), or fs.writeFile(). Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file is not accessible.

Wouldn't it be preferrable to let `fs.readFile` error out in case of a problem?

packages/firestore/package.json-e
ikusteu: Mac…happened to me when I was using grep or find or something.

packages/backup-restore/src/firestore/restoreService.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    return { ok: true, data: "OK" };
+  } catch (err: any) {
+    return { ok: false, message: err.message };
+  }
+}
+
+//* #endregion restoreScripts */
+
+//* * #region restoreUtils */
```
ikusteu, line 36: Write region comments in any way you like, but when using `// #region <regionName>` syntax, VSCode lets you minimise the region (by clicking an arrow on the left hand side, next to line numbers)

ikusteu (changes requested):
Before this can go in, please:
- [ ] address @silviot's comment about creating config dir if one exists
- [ ] remove `*.json-e` files
~~- [ ] check if babel is necessary (as I've not seen it used anywhere)~~ **edit:** Babel is there for testing...right?

ikusteu:
Additionally, a couple of comments to leave as todos:
- [ ] do wrap the tests in describe blocks as that's much more organised that way in the reporter (when there are 500-something unit tests)

Some suggestions for enhancements:
- [ ] Add option validations
- [ ] Add an option to configure a path (in config, or cli) to a dir containing backups (for both backup and restore functionality)

Some more extensive suggestions:

1. For the `package.json` requirement. I would suggest adding a `backup-restore` config file (required) just like `tsconfig.json` or `firebase.json`, etc... Then, the working project root can be determined by nearest available `backup-restore.config.json` (going up the directory tree). This would work for both small projects (single `package.json`) as well as rush monorepo (no root `package.json`)

2. One problem I've found is that the current functionality can support only one level of document nesting in firestore. To give an example: We have `organizations/{organization}/customers/{customer}` which is fine, and is supported here, but when you look at the bookings, there is `organizations/{organization}/bookings/{bookingsCustomer}/bookedSlots/{bookingEntry}`. This is not supported as `bookedSlots` would be lost (only organization and booking, as it's subcollection, would get picked up). To remedy this, you would first need a recursive-friendly JSON schema. I would recommend something like this:
```typescript
interface BackupDoc {
  [documentId: string]: {
    data: DocumentData;
    subCollections?: {
      [collectionId]: BackupDoc
    }
  }
}
```
How easy (or hard) that would be to achieve in terms of validations, I don't know. I know I had problems achieving this with TS, but I've tried it just now and it works.

Furthermore, there should also be functionality to support this. My suggestion would be to have two functions:
1. **Function A** to iterate over all documents (in a collection) and return their data, plus the result of **Function B** over their `subCollections`
2. **Function B** to iterate over all collections (at document node) and return the results of **Function A** over each collection 

Additionally, we could then accept a root node, where passing `/organizations/eisbuk/bookings/1234` as a root node would return:
```typescript
{
  organizations: {
    eisbuk: {
      subCollections: {
        bookings: {
          "1234": {
            name: "saul",
            /* rest of the data */
            subCollections: {
              bookedSlots: {/* a record of booked slots */}
            }
          }
        }
      }
    }
  }
}
```
This would omit all documents and collections in breadth and go depth-only until `organizations/eisbuk/bookings/1234` and from there on would pick up all breadth of documents and collections down to the deepest one.

Additionally, we could add a `maxDepth` param where all documents and collections in breadth would get picked up, but to a certain level of depth (where there would be no check for subCollections)

ikusteu:
Aand, on a final note, I really like what you did so far. It's a considerable amount of work and the code organisation (in-file as well as folder structure) is clear as ever 👍. Especially the setters and getters. I also really like the CLI utils' (commander and interactor) api.

delasous:
Notes for any and all reviewers:
1. The package / CLI interface now wraps this third-party library: [firestore-backup-restore](https://github.com/dalenguyen/firestore-backup-restore) which provides a reliable solution for exporting nested collections/documents as Ivan highlighted above - `organizations/{organization}/bookings/{bookingsCustomer}/bookedSlots/{bookingEntry}` will now be returned if it exists.
2. There are two caveats for using this package:
    1. there is a now a version mismatch of `firebase-admin` in other packages which needs to be addressed before "backup-restore" functionality can be written into a functions cron-job (to be written up and addressed in a pending enhancement ticket);
    2. Jest does not play well with the way this package imports `{ initializeApp } from "firebase-admin/app"`. This has something to do with its default module resolver not knowing how to resolve the `exports` keyword in the `firebase-admin` package.json. It is fixable, and it will also need to be resolved in the above-mentioned enhancement ticket, but I didn't have time to get into it now. Therefore, I have removed the few test files that remained that were testing CLI command-function invocation (the actual backup-restore functionality is thoroughly tested in the third-party package).
3. The CLI now no longer depends on a `configstore`. The options that were managed here are handled via default command options. See the updated Readme for notes on how this works / run `db-dump <cmd> --help` for descriptions. 
4. To test run the functionality:
    1. spin up the emulators & the admin app, and create test users & bookings, 
    2. `build-link` the package, 
    3. Add the eisbuk credentials `serviceAccount.json` with `db-dump projects:add <pathToServiceAccount.json>`
    4. Run `db-dump backup -e -p eisbuk localhost`

silviot (approved):

PR #473, opened 2022-05-11
Fix/slots mobile overflow

### LINKED ISSUE #466: The trash icon sometimes can be hidden in slots view
If a slot has many intervals (6 in my tests) different zoom levels might end up hiding the trash can icon, and potentially also the pencil icon.

Screenshot from 2022-05-10 12-49-05

Screenshot from 2022-05-10 12-49-56

ikusteu:
Fixes #466 
Additionally, user can scroll through intervals on `SlotCard` if using mobile or touchpad horizontal scroll. Neither scroll snap not slider implemented as that would require additional work, but can be implemented as an enhancement: if needed and if the style will remain such after Alex's mockup.

### DIFF
--- a/packages/client/src/components/atoms/SlotCard/SlotCard.stories.tsx
+++ b/packages/client/src/components/atoms/SlotCard/SlotCard.stories.tsx
@@ -5,14 +5,14 @@ import { SlotInterface } from "@eisbuk/shared";
 
 import SlotCard from "./SlotCard";
 
-import { baseSlot } from "@/__testData__/slots";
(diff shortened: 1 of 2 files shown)


fadwamahmoud:
Firefox desktop view
eisbuk

ikusteu:
> Firefox desktop view eisbuk

Could you please check again...I've applied the fix and it works for me in Firefox on MacOS Monterey

fadwamahmoud (approved):

PR #478, opened 2022-05-15
Update/multiple adult categories

### LINKED ISSUE #468: Change categories
The new list of categories:

- Pre-competitive adults (it: Pre-Agonismo adulti)
- Pre-competitive minors (it: Pre-agonismo ragazzi)
- Course adults (it: Corso adulti)
- Course minors (it: Corso ragazzi)
- Competitive (it: Agonismo)

#### Replace the category names/labels:
- [x] Rename the categories in `@eisbuk/shared` and apply to internal usage
- [x] Update the translations to recognise new category names

#### Provide the fallback as to not break the app even if data (in firestore) not updated:
- [x] when displaying the slots, should (temporarily) show all "adults" slots to "unsorted" adults
- [x] in customer form, show a disabled choice for "adults" -> that way an admin will have visual clue that the customer is "unsorted" adult, but won't be able to store new customers as being "unsorted"
- [x] firestore rules should allow updates with existing "adults" category (slots and customers), but disallow new customers or slots with said category

#### Create migrations:
Should create migrations for customers and slots, replacing the old "pre-competitive" and "course" categories with "pre-competitive-minors" and "course-minors" (as minors part was implied since we had a separate caegory for "adults")
- [x] Create migration for slots
- [x] Create migration for customers

ikusteu:
Fixes #468 

- new categories are created and the code updated (including forms/translations)
- there are fallbacks for deprecated category (as "adults" should be updated manually) and we want the app not to break if the admin hasn't yet ran the migration
- migration (function) is created to move the "pre-competitive" and "course" categories to "pre-competitive-minors" and "course-minors" respectively

for more info see #468


silviot (approved):
To be able to give this a go, I deployed the migration function to production.
While doing that I found a stale mention of `yarn` and replaced it with `rushx`.

@ikusteu please have a look at the commit I added.

PR #480, opened 2022-05-16
Enhancement/printable attendance

### LINKED ISSUE #472: Printable attendance changes
This is how the printable attendance should look like:

image

The top logo can be the organization name for now.

fadwamahmoud:
Fixes #472


silviot (changes requested):
This looks great!
There are a couple of changes needed though:

- ~~There should be no space between lines (`<hr>` should be removed)~~
- [ ] The first four columns should be narrower (as narrow as possible)
- [x] The column "Pista" (Ice Rink in Italian) should actually show the contents of te `Notes` field. That's where the actual rink will be specified. We can change the heading of the fourth column `Pista` → `Note`.
- ~~Let's make the column mentioned in the previous point (`Pista`, to be changed to `Note`) the one that has a different background color based on ice/off ice~~
- [x] The top "ice"/"Off ice" labels, next to the date, are not coloured when printed.  They should be.
- [ ] The vertical padding can be reduced. We should be able to fit about twenty rows on a single page.

fadwamahmoud:
> This looks great! There are a couple of changes needed though:
> 
>     * [ ]  There should be no space between lines (`<hr>` should be removed)
> 
>     * [ ]  The first four columns should be narrower (as narrow as possible)
> 
>     * [ ]  The column "Pista" (Ice Rink in Italian) should actually show the contents of te `Notes` field. That's where the actual rink will be specified. We can change the heading of the fourth column `Pista` → `Note`.
> 
>     * [ ]  Let's make the column mentioned in the previous point (`Pista`, to be changed to `Note`) the one that has a different background color based on ice/off ice
> 
>     * [ ]  The top "ice"/"Off ice" labels, next to the date, are not coloured when printed.
> 
>     * [ ]  The vertical padding can be reduced. We should be able to fit about twenty rows on a single page.

@silviot  Okay. But the <hr/> are there to separate slots, wouldn't it be confusing if they're all in one block?
Also if the ice and off-ice aren't colored on printing how would users know which color is which type?

silviot:
> @silviot Okay. But the `<hr/>` are there to separate slots, wouldn't it be confusing if they're all in one block?
I see I was probably looking at an extreme case where each athlete booked a single slot. So it was not really representative of the actual data we'll be likely to have. I take that back.

> Also if the ice and off-ice aren't colored on printing how would users know which color is which type?
I think we should have two coloured elements:
* the top `ice`/`off ice` symbols next to the date, that serve as a legend
* ~~The cells of the fourth column (instead of the first and second as we have now)~~ Actually, I think the way you did it makes more sense than what was in the sample. So I also take this back. We can revisit it if necessary.

silviot:
@fadwamahmoud Excellent! One more (hopefully small) thing:
the fifth column (insegnante) could be much narrower. The space freed by narrowing it should be used to reintroduce the rightmost column (Notes, so a duplicate heading of the fourth column - sorry if it's confusing) that is meant to host handwritten notes on paper (as opposed to the fourth column supposed to print data from the database).

silviot (approved):

silviot (changes requested):
@fadwamahmoud Excellent! One more (hopefully small) thing:
the fifth column (insegnante) could be much narrower. The space freed by narrowing it should be used to reintroduce the rightmost column (Notes, so a duplicate heading of the fourth column - sorry if it's confusing) that is meant to host handwritten notes on paper (as opposed to the fourth column supposed to print data from the database).

PR #506, opened 2022-06-02
Provide default customer id in new customer form

### LINKED ISSUE #503: The "New athlete" form should provide a sensible default for the "card number" field
A sensible default value can ba calculated this way:
* Filter all existing card numbers, taking only the ones that represent an integer
* Take the highest integer from previous step
* Add 1

silviot:
Fixes #503


ikusteu (approved):
LGTM,
I think the matrix is genius as well as the fact that it's only ran on a specific branch, by choice

Issue #514, opened 2022-06-06
Feature: TabItem: 2 pts


ikusteu:
## Component: TabItem

**Description:** A button used for tab-switching (in `customer_area` for instance). Different than the button as it should be more akin to the square shape, rather than inline button on smaller screen sizes.

**Mockup:**
[![Screenshot from 2022-06-07 12-47-00](https://user-images.githubusercontent.com/396425/172361765-1e914946-7fcb-4113-8f6c-dce0d3af7932.png)](https://www.figma.com/file/oAzgMQW7vCaZAYmL58lxeo/Eisbuk?node-id=1505%3A26404)

#### Props (should extend `React.HTMLAttributes<HTMLButtonElement>`):

- **Icon** (required: SVGComponent)
- **variant** (optional: string enum):
  - enum values:
    - text (default)
    - fill
  - alternatively, this could just be an `active` boolean, as I imagine it will be used in that capacity

#### Styles:

- We sould check the hover state mockup (my guess is just to make it like librocco navItem button, `color: white` on hover, rather than `gray-200`)

#### Storybook (stories):

- Variants

#### Functionality (should be tested):

- onClick

### Possible enhancements:

- null

ikusteu:
Fixed in #523

PR #523, opened 2022-06-09
Feature/tab item

delasous:
Implements #514


packages/ui/src/TabItem/TabItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+    </button>
+  );
+};
+
+const svgClasses = ["w-6", "h-6"];
+
+const buttonBaseClasses = [
+  "flex",
+  "flex-col",
+  "sm:flex-row",
```
ikusteu, line 44: We didn't specify breakpoints in the mockup (neither did Alex). I think this could go up to `md`, but let me know your opinion.

ikusteu: Aaand yeah...I've just checked the breakpoints and I think it's safe to move this to `768px +`. So that it's still square(ish) on mobile for easier button touch.

delasous: Ok, agreed - I will update

packages/ui/src/TabItem/TabItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+    ? buttonActiveClasses
+    : buttonDefaultClasses;
+
+  const svgClassNames = svgClasses.join(" ");
+  const buttonClassNames = [...buttonBaseClasses, ...buttonStateClasses].join(
+    " "
+  );
+
+  return (
+    <button className={buttonClassNames} onClick={onClickHandler} {...props}>
```
ikusteu, line 30: We should be able to pass `className` to the component (like so: `<TabItem {...props} className="absolute left-4" />` or smtn.) and those classes should be passed to a wrapping element (in this case `<button>`). This would mostly be used for positioning, min-w, etc.

ikusteu: I see now you collect all the `...props` and spread them into `<button>` element. In case of extending classes (as illustrated above), the custom `className` would override all of the default classes. Therefore classes should be extended (like so `const buttonClassNames = [...aBunchOfDefaultClasses, className].join(" ")`)

ikusteu (changes requested):
LGTM generally, just make sure you enable the extending of the container (`<button>` element) classes and take a look at the breakpoints discussion.

ikusteu:
Oh and you need to add `TabItem` export into `packages/ui/src/index.ts` so that the component gets built along with the `ui` package.

packages/ui/src/TabItem/TabItem.tsx
delasous: Ok, noted - I will update!

delasous:
> Oh and you need to add `TabItem` export into `packages/ui/src/index.ts` so that the component gets built along with the `ui` package.

Ok - this is also missing from UserAvatar. I will make sure to do it going forward

PR #524, opened 2022-06-09
Feature/slot type icon

### LINKED ISSUE #498: Feature: SlotTypeIcon: 1 pt
## Component: SlotTypeIcon
A simple touch-up to the existing `SlotTypeIcon.tsx`. New component should be written in `packages/ui`, and be global, as it will be used in Admin components when revamped. 

[**Mockup**:
Screenshot from 2022-06-07 12-50-51](https://www.figma.com/file/oAzgMQW7vCaZAYmL58lxeo/Eisbuk?node-id=1512%3A29283)

### Requirements:
Copy existing `SlotTypeIcon`, remove MUI components & styles, and add our own styles and svgs to match Figma mock

**Storybook**
- Default showing both variants: ice / off-ice

fadwamahmoud:
Fixes #498


packages/ui/src/SlotTypeIcon/SlotTypeIcon.tsx
```diff
@@ -0,0 +1,48 @@
...
+        <Icon />
+      </span>
+      <span key="type">{t(SlotTypeLabel[type])}</span>
+    </div>
+  );
+};
+
+const baseClasses = [
+  "px-4",
+  "py-3",
```
ikusteu, line 33: This is unnecessary (at least from the current state of the mockup)

ikusteu: Plus, it's easier to position the element if there's no weird padding (for instance when using `position: absolute`, I want to specify `left: 4px` and `bottom: 4px`, the icon itself won't appear the that position, but have a padding in between)

packages/ui/src/SlotTypeIcon/SlotTypeIcon.tsx
```diff
@@ -0,0 +1,48 @@
...
+  "text-base",
+  "text-bold",
+  "text-gray-500",
+  "flex",
+  "flex-row",
+  "items-center",
+  "justify-center",
+  "md:flex-row",
+  "md:px-4",
+  "md:py-3",
```
ikusteu, line 45: Default flex is `flex-row` so no need to specify it. If you have `px-4` and `md:px-4`.

ikusteu, line 45: No need to specify this (as I've mentioned above). Additionally, if you have `px-4` and `md:px-4` the `md:px-4` is completely unnecessary.

The way Tailwind's mobile first approach works is there's a default value (`px-4` in this case) and then the first breakpoint found is applied for all screen sizes above.

Example:
`px-4 sm:px-6 md:px-8` will use 
- Screen width: 0px - `sm`(excluding `sm`): `px-4` 
- Screen width: `sm` (including `sm`) - `md` (excluding `md`): `px-6` 
- Screen width: `md` (including `md`) and above: `px-8`

ikusteu (changes requested):
Please remove the excess padding, and take a look at the explanation for Tailwind responsive classes.

Additionally, the component should be exported from `packages/ui/src/index.ts` so that it gets built along with the `ui` package

ikusteu (approved):

Issue #512, opened 2022-06-06
Feature: UserAvatar: 2 pts


ikusteu:
## Component: UserAvatar

**Description:** An update on `EisbukAvatar` to conform to the new mockup

[**Mockup (avatar on the right hand side):**
![Screenshot from 2022-06-07 12-48-38](https://user-images.githubusercontent.com/396425/172362035-a884f322-2292-4ae2-9a39-8b129623b79e.png)](https://www.figma.com/file/oAzgMQW7vCaZAYmL58lxeo/Eisbuk?node-id=1514%3A33847)

#### Props (a subset of `CustomerBase` interface):

- **name** (required: string):
- **surname** (required: string):
- **avatar** (optional: string):
  - this isn't used currently and the `CustomerBase` interface should be extended with it
  - alternatively it can be `image` or `profilePicture`, etc.
  - in absence of this property (or in case of an empty string), some fallback should be shown
- **category** (required: string):
  - we've prevouisly used this to render different mask (border radius) for the image, but can be omitted if not necessary anymore

#### Styles:

- Category dependant mask ??

#### Storybook (stories):

- Default
- Categories (if mask will still be used)

#### Functionality (should be tested):

- null

### Possible enhancements:

- null

#### TODO:

- [x] extend `CustomerBase` with `avatar`
- [x] provide fallback if `avatar` not specified (we can reuse the initials logic)
- [x] finish the component

PR #525, opened 2022-06-10
Feature/user avatar

delasous:
Implements #512

### DIFF
--- a/packages/ui/src/UserAvatar/UserAvatar.tsx
+++ b/packages/ui/src/UserAvatar/UserAvatar.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 
 import { CustomerBase } from "@eisbuk//shared";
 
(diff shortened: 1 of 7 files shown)


packages/ui/src/UserAvatar/UserAvatar.tsx
```diff
@@ -0,0 +1,71 @@
...
+      </div>
+      <div className={avatarClassNames}>
+        {avatarUrl ? (
+          <img
+            className="rounded-full object-cover"
+            src={avatarUrl}
+            onError={({ currentTarget }) => {
+              currentTarget.onerror = null;
+              setAvatarUrl(undefined);
+            }}
```
ikusteu, line 40: Intersting...could you provide the resource for this (I was completely unaware of this)

ikusteu (commented):
LGTM, could you just provide some resource on the `onError` thing

ikusteu (approved):
LGTM, could you just provide some resource on the `onError` thing

packages/ui/src/UserAvatar/UserAvatar.tsx
delasous: Same here.

I found this [SO: Replace image source on error](https://stackoverflow.com/questions/34097560/react-js-replace-img-src-onerror), and the [MDN Docs: error event](https://developer.mozilla.org/en-US/docs/Web/API/Element/error_event). It is apparently a global element attribute.

ikusteu (changes requested):
The component should be exported from `packages/ui/src/index.ts` so that it gets built along with the `ui` package

ikusteu (approved):

Issue #487, opened 2022-05-30
Feature: ActionDialog: 3 pts


delasous:
## Component: ActionDialog

**Description:** A dialog that will be used across the app as confirmation dialog. In case of `customer_area`, it will show a simple variant of #495 (details of the interval), and two buttons: to either reject or confirm the action.

Existing `ConfirmDialog` can be used as a reference to build atop.

**Mockup:**
[![Screenshot from 2022-06-08 12-11-03](https://user-images.githubusercontent.com/396425/172591706-3bfdd16b-099e-4d4c-b888-c960385df818.png)](https://www.figma.com/file/oAzgMQW7vCaZAYmL58lxeo/Eisbuk?node-id=1514%3A36896)

#### Props:

- **title** (optional: string):
  - if not provided, only the message (**children**) is shown
- **children** (optional: JSX.Element):
  - some simple prompts might only feature the **title** so this is optional
  - this is here so that we can pass some "content" to the dialog (as in the case of booking confirmation where the simple `IntervalCard` is shown)
- **confirmLabel** (optional: string):
  - label on confirm button
  - if not provided should fall back to something like "confirm" or "yes"
- **cancelLabel** (optional: string):
  - label on cancel button
  - if not provided should fall back to something like "cancel" or "no"
- **open** (required: boolean)
- **onClose** (required: Function): click handler used to close the modal in some outer state
- **onConfirm** (required: Function): handler for confirming action

#### Storybook:

- Default: we should provide some title and use the default values for **confirmLabel** and **cancelLabel**
- ...(variants):
  - this isn't a constrained choice of variants, but rather usage of this component "in the wild"
  - initially it should feature booking confirmation, but get extended with more examples as the mockup evolves

#### Functionality:

- should close itself on cancel
- should call to **onConfirm** and close itself on confirmation

#### Depends on:

- [ ] #486
- [ ] #495 (doesn't depend directly, but would be good to wait for it for stories)

#### TODO:

- [ ] Create prop interface confirming to the spec and TDD the functionality
- [ ] Replace MUI `Dialog`, `DialogTitle`, `DialogContent`, `DialogActions` with our tailwind styles to Figma mock.
- [ ] Replace MUI `Button` with our own local (`ActionDialog/Button.tsx`), which conforms to [this mock](https://www.figma.com/file/oAzgMQW7vCaZAYmL58lxeo/Eisbuk?node-id=1520%3A27766), and captures hoverable state
- [ ] Create stories

[...]

PR #528, opened 2022-06-12
Feature/action dialog

delasous:
Implements #487

### DIFF
--- a/packages/ui/src/ActionDialog/ActionDialog.stories.tsx
+++ b/packages/ui/src/ActionDialog/ActionDialog.stories.tsx
@@ -9,18 +9,17 @@ export default {
 } as ComponentMeta<typeof ActionDialog>;
 
 export const Default = (): JSX.Element => (
-  <div className="w-screen h-screen bg-gray-800">
-    <div className="absolute top-1/2 left-1/2 transform -translate-x-1/2 -translate-y-1/2 z-10">
-      <ActionDialog
-        title="Cancel booking for"
-        cancelLabel="Go Back"
-        confirmLabel="Confirm Cancellation"
-        onCancel={() => {}}
-        onConfirm={() => {}}
-      >
-        This is a placeholder message for a Simple Interval Card, but also stunt
-        doubles as a TextContent variant?
-      </ActionDialog>
-    </div>
+  <div className="absolute top-0 right-0 bottom-0 left-0 bg-gray-800">
+    <ActionDialog
+      className="absolute top-1/2 left-1/2 transform -translate-x-1/2 -translate-y-1/2 z-10"
+      title="Cancel booking for"
+      cancelLabel="Go Back"
+      confirmLabel="Confirm Cancellation"
+      onCancel={() => {}}
+      onConfirm={() => {}}
+    >
+      This is a placeholder message for a Simple Interval Card, but also stunt
+      doubles as a TextContent variant?
+    </ActionDialog>
   </div>
 );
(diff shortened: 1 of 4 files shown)


ikusteu (changes requested):
Add export for the component from root level `index.ts` in `ui`

packages/ui/src/ActionDialog/ActionDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <div className="flex space-x-2">
+        <Button
+          className="text-gray-700 font-medium bg-gray-100 hover:bg-opacity-0"
+          size={ButtonSize.LG}
+          onClick={handleCancel}
+        >
+          {cancelLabel}
+        </Button>
+        <Button
+          className="text-red-700 bg-red-200 hover:bg-red-100"
```
ikusteu, line 48: `text-<color>-<shade>` doesn't work here (thanks to button component's default `text-white`). We can easily override this by doing `!text-gray-700` (making in `!important`).
There definitely is a cleaner way of doing this, but I think, in this case, we can get away with using `!important` directive.

packages/ui/src/ActionDialog/ActionDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  return (
+    <div className="flex flex-col max-w-sm items-start gap-y-4 p-5 rounded-md bg-white">
+      <div>
+        <h3 className="text-base leading-6 font-semibold text-red-700">
+          {title}
+        </h3>
+      </div>
+      <div>{children}</div>
+      <div className="flex space-x-2">
```
ikusteu, line 39: I would add `w-full` and `justify-right` as that seems more natural for me visually, but do your judgement here.

packages/ui/src/ActionDialog/ActionDialog.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import ActionDialog from "./ActionDialog";
+
+export default {
+  title: "Action Dialog",
+  component: ActionDialog,
+} as ComponentMeta<typeof ActionDialog>;
+
+export const Default = (): JSX.Element => (
+  <div className="w-screen h-screen bg-gray-800">
```
ikusteu, line 12: Sorry about giving an opinion just now, but I've thought about what you mentioned (adding overlay). I think we can include the overlay in the component itself and have a click outside the modal body call to `onCancel`, thus closing the modal.
For this purpose I've copied the `useClickOutside` hook over to `ui` package and it can be used for this purpose.
For usage/reference, check [this](https://github.com/eisbuk/EisBuk/blob/master/packages/client/src/react-redux-firebase-auth/ui/atoms/AuthErrorDialog.tsx)

ikusteu (changes requested):
I left some comments for small fix-ups and a simple overlay enhancement (don't forget the root `index.ts` export)

packages/ui/src/ActionDialog/ActionDialog.stories.tsx
delasous: I think I made this comment before learning that the Dialog controller should only render one Dialog in a fixed position at a time. Is it better for the controller to define the overlay and use the `useClickOutside` hook, rather than each Dialog layering an overlay on top of the others?

ikusteu: Yeah...ok, it makes sense to make it just a stateless component. However, as such, it shouldn't care if it's a modal or not, so it can simply take in `onCancel` and `onConfirm` (can just be passed as click handlers) and the controller will decide if it wants to use it as modal or not, implement the concept of open/closed, etc.

ikusteu (approved):

PR #531, opened 2022-06-20
Fix: User Avatar Styles

delasous:
Add small screen right text-align


delasous:
Unsure why storybook build failed here? It worked locally

PR #546, opened 2022-06-25
Feature/notification store setup

### LINKED ISSUE #535: Feature: Notification data flow: 5 pts
## Data flow: Notifications

**Description:** This ticket should create a data flow for Notification logic. It should work similarly to Snackbar notifications we're already using and be consumed by the NotificationController component to render notifications in predefined place. It should utilise Redux for this (must include its own reducer, actions and selectors), thus enabling the controller component to read the state and different parts of the app to update.

### Requirements:

- only one notification can be shown at the time (as there's no space in the mock for multiples), so the store should house a queue, allowing the controller to read one entry at the time.
- queue should be used as FIFO ("pushing" to the end when adding and "unshift"-ing from the start when reading)

#### Types:

```typescript
enum NotificationAction {
  // We're stepping away from the standard "@@Eisbuk/ENQUEUE_NOTIFICATION"
  // as that would produce conflicts with Enqueue snackbar (while our app is in in-between state of rehaul).
  // Additionally, we might remove this as a separate package so it would make sense to namespace it so.
  Enqueue = "@@NOTIFICATION/ENQUEUE",
  Next = "@@NOTIFICATION/NEXT",
  Evict = "@@NOTIFICATION/EVICT",
}

interface NotificationEntry {
  // TBD
}
```

#### Store State (schema):

- **storeSchema**:
  - queue:
    - a list (array) of `NotificationEntry` to be used as FIFO (queue)
    - the first element in the queue (at index `0`) is the notification currently shown
    - if the queue is empty, no notification is shown

#### Actions:

- **enqueueNotification**:

  - thunk: should accept notification and dispatch action to the store
  - action:
    - type: `NotificationAction.Enqueue`
    - payload: `NotificationEntry`
  - reducer:
    - should push provided notification to the end of the queue
    - be careful to return new state rather than mutate the old one when working with **queue**

- **nextNotification**:

  - thunk: should dispatch action to the store
  - action:
    - type: `NotificationAction.Next`
  - reducer:
    - should check the **queue** for notifications other than the first (active one) and if additional notifications exist (`notifications.length > 1`), should remove the first element (making the next notification an active one)
    - if there are no more notifications in the queue, shouldn't do anything
    - be careful to return the new **queue** state rather than mutate the old one

- **evictNotification**:

  - thunk: should dispatch action to the store
  - action:
    - type: `NotificationAction.Remove`
  - reducer:
    - should remove the first element from the notifications **queue**:
      - if there are additional notifications in the queue, the next one will automatically be active
      - if the queue is empty, the notification is just removed (no notification is shown)

#### Selectors:

- **getCurrentNotification**:
  - should subscribe the notification store and derive the first (active) element of the notifications **queue**

#### Functionality:

To put everything (specified above) together:

- the store would house the **queue** of notifications where the first notification is the one rendered to the screen (active one)
- when adding new notification (using **enqueueNotification**), it would add the new notification to the end of the array
- the active notification (at `0` index of the queue) would be subscribed to by the consumer and received upon update
- when the consumer is ready to take in the next notification, it would perform **nextNotification**
- **nextNotification** would remove the first (active) notification from the **queue**, only if there are additional notifications waiting, resulting in making the next notification an active one, if there are no notifications waiting, np changes would be made
- when the consumer wants to stop showing the current notification, it would call **evictNotification**, which would always remove the notification from the start of the queue where the next one (if exists) would automatically become an active one, or (if there are no other notifications in the queue) no notification would be shown

#### Tests:

- all of the functionality should be tested (against an instance of the real store, generated using `getNewStore`)

### Possible enhancements:

- add an additional state to the **notifications** store:

  - **canEvict**:
    - boolean
    - a state that should be `false` when the new notification becomes active (arrives at the start of the queue)
    - should become `true` if **nextNotification** is dispatched to the store and there were no additional notifications in the queue at the time
    - essentially (if true) it means: "this notification wasn't yet removed, but it can be if necessary (if the new one comes in)"

- update the reducer to take **canEvict** into account:
  - when performing **readNotification** should set it to `true` if the active notification hasn't been removed from the queue (no other notifications in the queue)
  - when enqueueing notification (if `true`, should immediately replace the active notification with the new one)
  - when removing an active notification from the queue (essentially making the next notification a new one), should set to `false`

### Progress:

- [ ] Define the types (`NotificationEntry`)
- [ ] Set up initialState
- [ ] Develop the actions/reducer logic (preferably using TDD)
- [ ] Add the selector
- [ ] Test the logic through and through (if not developed using TDD)

ikusteu:
Fixes #535 
including the enhancement mentioned in the ticket.

### DIFF
--- a/packages/client/src/store/actions/__tests__/notifications.test.ts
+++ b/packages/client/src/store/actions/__tests__/notifications.test.ts
@@ -7,14 +7,15 @@ import {
   nextNotification,
 } from "../notificationsActions";
 
-const testNotification1 = {
-  variant: NotifVariant.Success,
-  message: "We tha best music!!!",
-};
-const testNotification2 = {
-  variant: NotifVariant.Error,
-  message: "U can't touch this",
-};
+// #region tableTestRunner
+interface TestParams {
+  name: string;
+  testQueue: Omit<NotificationInterface, "key">[];
+  testCanEvict: boolean;
+  run: (store: ReturnType<typeof getNewStore>) => void;
+  wantQueue?: Omit<NotificationInterface, "key">[];
+  wantCanEvict?: boolean;
+}
 
 const setupQueue = (
   store: ReturnType<typeof getNewStore>,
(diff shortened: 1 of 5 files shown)


ikusteu:
This work is included in #549

ikusteu:
This work is included in #549

PR #558, opened 2022-06-28
Feature/use notification

### LINKED ISSUE #536: Feature: useNotifications hook: 8 pts
## Data flow: Notifications

**Description:** This ticket should create a hook controlling the notifications [store](https://github.com/eisbuk/EisBuk/issues/535). The notifications entry in Redux store is envisioned as holding state and providing actions to update the state in orderly manner (enqueueing and reading from queue), but the hook should be in charge of using those actions to control the flow of displaying notifications from the queue in Redux store.

_There should also be a third component to this flow - the NotificationController (#519) itself, which will be in charge of rendering notifications themselves and providing animations. Even though I don't imaine this hook being used outside of said component (and the hook logic could just be written inside the component then), I think it makes sense still, to separate the concerns here as this hook could, as an enhancement at some point, be merged with the Store side: using subscriptions inside of thunks so that the store "updates itself" and just notifies the component of the updates._

### Requirements:

- only one notification can be shown at the time (as there's no space in the mock for multiples), so the hook should be in charge of returning the current value of subscribed active notification (the first one in the queue), "requesting" next item (if any enqueued), using **nextNotification** (as specified in #535) and removing the currently active notification from the **queue** using **evictNotification** after a certain timeout.
- important thing to note here are the timeouts:
  - **minTimeout**:
    - the minimum amount of time the notification should be shown
  - **maxTimeout**:
    - the maximum amount of time the notification should be shown
  - both should be configurable
  - after the new notification is shown (active notification changes), the hook should try and perform **nextNotification** after **minTimeout**, to allow for enqueued notifications (if any exist) to flow frequently, and if there are no changes in active (meaning **nextNotification** didn't find any additional notifications in **queue**), after **maxTimeout**, the notification should be removed (using **evictNotification**)
- user should also be able to remove notification on click of a button, so the hook should also return **handleRemoveNotification**, which would, under the hood dispatch **evictNotification** to the store.
- circling back to timeouts, there should be some internal store for timeouts (**useRef** would be preferred IMO), so that the timeouts can be cancelled if something new comes in (so that stale timeout callbacks don't run when we no longer want them to, causing weird and unpredictable behaviour).

#### Types:

```typescript
interface NotificationEntry {
  // TBD
}
```

#### Hook interface:

- **Parameters:**

  - **timeouts** (optional: object):
    - **minTimeout** (optional: number):
      - minimum time the notification should be shown for (in milliseconds)
      - should account for entry animation so that the notification has time to enter the screen and be read by the user
      - should provide some sensible fallback if not specified
    - **maxTimeout** (optional: number):
      - maximum time the notification should be shown for (in milliseconds)
      - we're doing this for notifications to not linger on the screen, but be removed after a while (even if there aren't any more enqueued), as described above :point_up:
      - should provide some sensible fallback if not specified

- **Returns:**
  - **result** (object):
    - **active** (`NotificationEntry`):
      - the currently active notification entry (first one in the queue, should be rendered to the UI)
    - **handleRemoveNotification** (Function):
      - a handler to be called if user wants to manually remove the notification from the screen (as described above :point_up:)

#### Functionality:

To put everything (specified above) together:

- the hook would be initialised with **minTimeout** and **maxTimeout** (either specified by the user, or assigned from fallback)
- the hook would subscribe to `notifications` entry in Redux, receiving from it the active notification (if any)
- each time the active notification changes, the hook should set two timeouts, using `setTimeout` (**afterMinTimeout** and **afterMaxTimeout**, naming can be different here) and store both to respective refs:
  - after a **minTimeout** has passed, the **nextNotification** action should be dispatched (that should be handled by **afterMinTimeout** callback)
  - if there aren't any notifications in queue, no change will take place
  - if no changes took place, the notification should be removed after **maxTimeout** has passed (that should be handled by in **afterMaxTimeout** callback)
  - each time the active notification changes, before setting any timeouts, a check should be performed to see if there are any timeouts left from the cycle before and if so should clear both (to prevent unwanted updates to the store)
- the hook should always return currently active notification and **handleRemoveNotification**, which should, on call, dispatch **evictNotification** to the store

#### Tests:

- ~~all of the functionality should be tested against an instance of the real store, generated using `getNewStore`. Additionally, `testHooksWithRedux` could be used. It should just be modified slightly to return the return structure of the hook.~~ - as @silviot pointed out, this should probably be tested using cypress (but after the entire functionality is implemented)

### Possible enhancements
- make this into a context `Provider`, rather than a hook, as it might be consumed at different places at the same time (and we don't want two instances manipulating an ordered queue at the same time) - as specified in #551

fadwamahmoud:
Fixes #536


packages/client/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,54 @@
...
+  console.log({ active });
+  const afterMinTimeout = useRef<NodeJS.Timeout | null>(null);
+  const afterMaxTimeout = useRef<NodeJS.Timeout | null>(null);
+
+  /**
+   * Possible bug: when I click next when it's exiting/other one is entering it's skipped
+   */
+  useEffect(() => {
+    afterMinTimeout.current && clearTimeout(afterMinTimeout.current);
+    afterMaxTimeout.current && clearTimeout(afterMaxTimeout.current);
```
ikusteu, line 40: Ok, so this is good enough, but...this works in most cases, not in one.
It works to prevent timeouts firing against wrong notification ✅ 
It won't clear timeout on dismount:
- the timeout is set
- in the mean time (before timeout ran out) the hook is dismounted (the component calling a hook is dismounted)
- the timeout runs out and fires a callback

This shouldn't be a problem in this use case, but:
- it represents a memory leak
- would in very similar cases (this exact case, but having multiple components consuming/depending on the store) produce unexpected behaviour: the function running after it should have been cleaned up (on dismount)

ikusteu: Additionally, your case could be written like so:
```typescript
useEffect(() => {
  afterMinTimeout.current = setTimeout(() => {
    dispatch(nextNotification());
  }, minTimeout);

  afterMaxTimeout.current = setTimeout(() => {
    dispatch(evictNotification());
  }, maxTimeout);

  return () => {
    afterMinTimeout.current && clearTimeout(afterMinTimeout.current);
    afterMaxTimeout.current && clearTimeout(afterMaxTimeout.current);
  };
}, [active]);
```
This way, the `useEffect` is ran at mount. When the `active` changes, it should be ran again, however, before it is ran again, the cleanup function is ran first, clearing timeout, only then is the `useEffect` block ran again, setting up new timeouts, and rinse and repeat for each change of `active`. Finally, when the hook is dismounted, it will run a cleanup function, clearing all (if any) outstanding timeouts.

packages/client/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,54 @@
+import {
+  nextNotification,
+  evictNotification,
+} from "@/store/actions/notificationsActions";
+import { getActiveNotification } from "@/store/selectors/notifications";
+import { NotificationInterface } from "@/types/store";
+import { useEffect, useRef } from "react";
+import { useDispatch, useSelector } from "react-redux";
+
+interface useNotificationsParams {
```
ikusteu, line 10: It is a convention to write the types, interfaces and enums in **PascalCase**. This is to prevent from clashing with JS objects maybe named the same way, only in **camelCase**

packages/client/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,54 @@
...
+  timeouts: {
+    minTimeout?: number;
+    maxTimeout?: number;
+  };
+}
```
ikusteu, line 15: This also works, don't get me wrong, but it's ok to lift is one level:
instead of having 
```typescript
params: {
  timeouts: {
    // min and max timeout
  }
}
```

we can simply use 
```typecript
timeouts: {
  // min and max timeout
}
```

packages/client/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,54 @@
...
+
+interface useNotifications {
+  active: NotificationInterface;
+  handleRemoveNotification: () => void;
+}
```
ikusteu, line 20: 1. This isn't used anywhere, aside from accidental export, a result of clashing of same named interface and a hook (as mentioned above ☝️ )
2. You can use and interface to declare a function, but it should be written like this:
```typescript
interface Function {
  (<params>): <return type>
}
```
and then you can just declare a function like so: `const function: Function = (params) => {/* ... */}`, without having to write the types for params and return again

ikusteu (changes requested):
All in all, a to-the-point solution, with that `useEffect` cleanup tweak. 

Also, please clean up the code (the comments regarding type declarations) and do remove the `console.log` and a comment about the functionality being buggy: it works exactly as it's supposed to 👍, the story is faulty - it's in its intermediate state and I will update it to adhere to final functionality

packages/client/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,56 @@
...
+  const active = useSelector(getActiveNotification);
+  const handleRemoveNotification = () => {
+    dispatch(evictNotification());
+  };
+  const afterMinTimeout = useRef<NodeJS.Timeout | null>(null);
+  const afterMaxTimeout = useRef<NodeJS.Timeout | null>(null);
+
+  useEffect(() => {
+    afterMinTimeout.current && clearTimeout(afterMinTimeout.current);
+    afterMaxTimeout.current && clearTimeout(afterMaxTimeout.current);
```
ikusteu, line 37: This can be removed as the cleanup already does it

packages/client/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,56 @@
...
+interface Timeouts {
+  minTimeout?: number;
+  maxTimeout?: number;
+}
+interface useNotificationsHook {
+  (timeouts: Timeouts): {
+    active: NotificationInterface | undefined;
+    handleRemoveNotification: () => void;
+  };
+}
```
ikusteu, line 21: Good job, but make it upper case

ikusteu (changes requested):
Good job, just two small edits to take care of.

ikusteu (approved):

Issue #559, opened 2022-06-29
Enhancement: Modal


ikusteu:
Upgrade the modal feature functionality to enable the modal dialogs to stack on top of each other.

- [ ] use the same logic (type interfaces, whitelisted components open/close logic), but apply it to a stack of modal elements rather than a single one:
- the state should instead of current `ModalState` be `ModalState[]` (and the interface should probably be renamed to `ModalElement`)
- the array should serve as a stack (LIFO) so that new modals are stacked on top of each other and are closed in reverse order
- opening a new modal should push to the stack
- closing a modal should pop from the stack (closing an element somewhere in the middle of stacked modal elements is not possible)

- [ ] the component should return `null` (not render anything) if no modal elements on stack

PR #560, opened 2022-06-29
Feature/modal

### LINKED ISSUE #542: Feature: CancelBookingDialog
## Component: CancelBookingDialog

**Description:**
This is a controller component used to render `ActionDialog` with confirmation to cancel booking. As a `controller` component it is part of the app layer and as such it is hooked to the app state.
It should receive the slot/booking data and render a `simple` variant of an `IntervalCard` component as `ActionDialog` content.

**Mockup:** [Figma](https://www.figma.com/file/oAzgMQW7vCaZAYmL58lxeo/Eisbuk?node-id=1514%3A36896)

### Requirements:

- components:
  - #487
  - #495

#### Props (should extend `SlotInterface`):

- **interval** (required: date string):
  - an interval to render
- **onClose** (required: Function):

  - a part of standard modal related props
  - a function dispatched when the modal should close

#### Storybook (stories):

- This is an app layer component and as such doesn't get much from storybook

#### Functionality (should be tested):

- should call `onClose` on both `onConfirm` and `onCancel`
- should dispatch `cancelBooking` on confirmation

### Possible enhancements:

- null

### LINKED ISSUE #543: Feature: Modal
## Component: ComponentName

**Description:**

A singleton modal controling component. It should be able to receive the variant and props from Redux and render the appropriate component with appropriate content.

**Mockup:**
No mockup per se. The [Action Dialog](https://www.figma.com/file/oAzgMQW7vCaZAYmL58lxeo/Eisbuk?node-id=1514%3A36896) can be used as a reference, with an overlay, behind the dialog, covering the rest of the screen, so that the rest of the screen appears faded with respect to the dialog in modal.

### Requirements:

The rendering of dialog components is envisioned as rendering a specific component and passing the props.

- the components themselves should be created (as controller components) in a different file (we can use `controllers` folder to house those).
- each controller component should be able to receive two sets of props:
  - for its specific use case (if rendering an `ActionDialog` with an `IntervalCard` it should receive the data for a slot for which to render a confirmation dialog)
  - some standard modal related props such as `onClose` (this should be standardised across modal controller components)
- the `Modal` component should have a lookup connecting each **variant** with appropriate controller component

#### Elements:

- Overlay:
  - an overlay on top of the rest of the page, fading the contents
- Modal content:
  - an appropriate component rendered "inside the modal"

#### Store:

- **name:** modal
- **state** (following structure :point_down: or `null`):

  - **variant** (required: string enum):
    - CancelBooking:
      - component: #542
    - (...):
      - variants should be extended as the use cases for modal evolve
  - **content** (required: `Record<string, any>`):
    - the content structure will be appropriate for each variant
    - :point_up: this is doable with typescript, by specifying te expected structure per each variant. That will provide type safety when consuming the content, as well as having TS yell at us if we want to update the store with mismatching content (or possibly mismatchnig content, TS is a Nazi for that stuff)

- **actions:**
  - **closeModal:**
    - a simple action setting the **modal** state to `null`
  - **openBookingConfirmation:**
    - should receive slot/booking information as an argument and open a modal with `CancelBooking` variant
  - **(...):**
    - additional actions should be added for each modal use case

#### Storybook (stories):

- null (This is an app layer component and states would be hard to reproduce in storybook)

#### Functionality:

- the component should render using [React.createPortal](https://reactjs.org/docs/portals.html) into a newly created `div` with `id="modal"`
- the component should create the aforementioned `div` element on mount and perform a check if the element already exists (to avoid adding multiple container elements)
- the fact that the **modal** portion of the store is defined (and non-null) should be processed as modal being `open` and, of course, the **modal** in store being undefined would mean the modal is closed

### Possible enhancements:

There will be a use case requiring multiple modals to be stacked on top of each other. This is not necessary ATM and can be an enhancement. I that case, I guess the modal should receive an array of variant/contents.

#### Depends on:

- #542: in order to implement at least one variant

#### TODO:

- [ ] Create a type logic so that each variant is connected to its specific content structure (this will be used in thunks, for checking, as well as to infer types in the `Modal` component)
- [ ] Create a Redux entry for a **modal** and provide appropriate actions
- [ ] Create a modal component (using react portal and open/close logic)
- [ ] Create a **variantComponent** lookup determinig the appropriate controller component to render for each variant (currently, only the `cancel-booking` variant is defined)

ikusteu:
Fixes #542 
Fixes #543 
Enhancement for #543 is in #559


packages/client/src/features/modal/types.ts
```diff
@@ -0,0 +1,48 @@
+import componentWhitelist from "./components";
```
silviot, line 1: I can't find the definition of `componentWhitelist`. What am I missing?

packages/client/src/features/modal/components/index.ts
```diff
@@ -0,0 +1,3 @@
+import CancelBookingDialog from "./CancelBookingDialog";
```
silviot, line 1: I expected to find the definition of `componentWhitelist` here but it's not here. What am I missing?

ikusteu: 👇 the default export: same as writing `const componentWhitelist = {...}; export default componentWhitelist`

packages/client/src/features/modal/types.ts
```diff
@@ -0,0 +1,49 @@
...
+ * each time.
+ */
+type WhitelistedComponents = keyof typeof componentWhitelist;
+
+/**
+ * This contraption returns appropriate prop interace for specified
+ * component. Provided that the component is in the whitelist of components
+ * that can be rendered inside the Modal
+ */
+type GetComponentProps<C extends WhitelistedComponents> = Omit<
```
delasous, line 19: Smooth moves!

delasous (approved):
Really nice work!

packages/client/src/features/modal/types.ts
ikusteu: 😉

PR #562, opened 2022-06-29
Invisible buttons

### LINKED ISSUE #555: Buttons are unreadable: white on white
See https://storage.googleapis.com/cypress-logs-bucket/2022-06-27/2571632958/storybook-client/index.html?path=/story/unauthorized-page--default

ikusteu:
Fixes #555 
Add a workaround to fix "invisible" buttons due to Tailwind and Mui clashing:
- Tailwind features a preflight reset css file removing background colour on buttons
- PostCSS, for some reason loads generated CSS files (including the preflight) as latest - highest priority
- button background for primary button gets overwritten by reset

- A "quick" fix - assign colour to buttons explicitly through class names: classes created with `useStyles` are created in the browser so they have the highest priority (are loaded latest)

Not ideal, but we're moving to Tailwind bit by bit anyhow and it's good enough to make it work at the moment.


delasous (approved):
LGTM

PR #564, opened 2022-06-30
Fix: Layout Notifications

delasous:
Quick fix for z-index of Notifications container. Tailwind's max pre-defined z-index is "z-50"


ikusteu (approved):
Lol...that was fast

PR #567, opened 2022-07-03
Rehaul/customer area

### LINKED ISSUE #540: Page: CustomerAreaPage
## Page: CustomerAreaPage

**Description:** A page component to be used for `/customer_area` (booking) route.

**Mockup:** [Figma](https://www.figma.com/file/oAzgMQW7vCaZAYmL58lxeo/Eisbuk?node-id=1503%3A26254)

### Requirements:

Should use the components created in this sprint to render the customer area page to mock.
Should read and display data from store:

- customer (current customer)
- slots available for booking (with respect to customer's category)
- booked slots

Unlike the old page component, it seems that the split between ice/off-ice slot is not necessary (they will be visually different using IntervalCard styles)

Previous version also featured subroutes for `book-ice`, `book-off-ice` and `calendar`, but I feel this can just be local state.

#### Elements:

- Layout:
  - components:
    - #500
    - #493
    - #519
    - #496
  - The page layout comprised of standard `Layout` component and `CalendarNav`
  - The `Notifications` part of the layout component should be `NotificationContainer` component (hooked to the store and in charge of it's own updates)
  - Calendar nav should dispatch the date change to redux
- Days:
  - components:
    - #495
    - #537

#### Props:

- No props, all data should be read from the app (Redux) store

#### Storybook (stories):

- null

#### Styles:

- the content should utilize the `content-container` class to position everything centrally

#### Functionality (should be tested):

- Should switch date (dispatch against Redux)
- Should book/cancel intervals (dispatch against Redux)

#### Depends on:

    - #500
    - #493
    - #519
    - #496
    - #495
    - #537

#### TODO

- [ ] Set up components on the screen (styles and all)
- [ ] Set up local state with `Book` and `Calendar` buttons
- [ ] Check the selector logic from the old page component (for slots/bookings, etc.)
- [ ] Write simple tests (book, date change)
- [ ] Check for additional tests (in the old page)

ikusteu:
Fixes #540 
The old one is still in the codebase in `pages/customer_area_archive` so that it's easier to pin point artefacts (functions, components, types, etc...) used only by the old page and replaced/removed in the now one.
The old page is also ignored by TS, jest and eslint so there should be no problems and the ignore patterns should be removed when the old page has been remove altogether.


packages/client/src/pages/customer_area/views/Book.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  const { handleBooking, handleCancellation } = useBooking();
+
+  return (
+    <>
+      <BookingsCountdownContainer />
+
+      {daysToRender.map(({ date, slots }) => (
+        <SlotsDayContainer key={date} date={date}>
+          {slots.map(({ interval, ...slot }) => (
```
delasous, line 27: Should slots first be sorted by their first intervals start time?
The intervals within each `IntervalCardGroup` are ordered by duration, but these groups are often not in chronological order - see screens below.

Screen Shot 2022-07-05 at 12 05 11 PM
Screen Shot 2022-07-05 at 12 08 11 PM

delasous (approved):
LGTM.

packages/client/src/pages/customer_area/index.tsx
```diff
@@ -1,142 +1,130 @@
...
-        ? getSlotsForCustomer(customerData.category, timeframe, date)
-        : () => ({}),
-    [date, customerData, timeframe]
+  return (
+    <Layout
+      Notifications={NotificationsContainer}
+      additionalButtons={additionalButtons}
+      user={customerData}
+    >
+      <CalendarNav {...calendarNavProps} jump="month" />
```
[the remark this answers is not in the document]
ikusteu: Yeah...I was thinking of including it it the `Layout` component and render conditionally (similar to the idea for admin appbar).

But thanks for reminding me of this. I'll note it down for enhancement.

packages/client/src/pages/customer_area/views/Book.tsx
ikusteu: I share your "frustration" with the way these things are...and am also confused as to how to go about it...should we just remove the `BookingCardGroup`* and have the intervals ordered in a way that the longest are first (regarding of which slot they belong to) or order the slots in ascending order and then apply the ordering within (in this case I'd keep the `IntervalCard` group). I've already noted this for enhancement and would like this to go in ASAP so that Fadwa can start on e2e while someone else (me) works on enhancements/fixes.

*\*it isn't all too necessary without the faded logic and serves only as a minor abstraction to keep things cleaner*

PR #576, opened 2022-07-06
Enhancement/admin bar

### LINKED ISSUE #565: Anhancement: Layout - add admin bar
I've been thinking about the admin view on layout and there's been some talk of admin nav. This could probably be a part of Layout.tsx or a separate file (in Layout folder) imported into the Layout. 

It should not render for a regular customer, but be shown for an admin.

#### Props (including existing props, plus adding the following):
- **isAdmin** (optional: boolean):
  - would control displaying of the admin bar
- **adminLinks** (array of items):
  - this can always be passed (regardless of **isAdmin** being `true` or not) as the nav bar wouldn't be shown at all if not **isAdmin** 
  - each item:
     - Icon
     - label
     - slug
     

#### Elements (for admin bar):
- **container:**
  - I've imagined a container as a navigation row (same height as the rows in current layout's header) and be added on top of everything.
  - same background colour as the rest of the header
- **admin nav:**
  - **Desktop:** 
    - in my head, it would look good positioned to the rightmost side (following the same padding/`content-container` logic as subsequent rows
    - my idea is to make the links a button group looking something like a group of `TabItem`s connected together and being rounded altogether (a bunch of square buttons with no margin in between) and have the container element utilise `overflow-hidden` and `rounded-<some-radius>`). To terribly "sketch" it out: `(button|button|button|button)` where `|` is straight edge, not margin and `()` are the mildly rounded edges of the button group
  - **Mobile:** 
    - a hamburger button with expanded menu overlaying the entire screen.
    - expanded menu could be made the same colour (background and text) as the header, or a contrast (grey text on white bg)
    - positioned as `fixed`, top, left and right `0`

fadwamahmoud:
Fixes #565


packages/client/src/features/modal/components/Modal.stories.tsx
```diff
@@ -19,7 +19,7 @@ const store = getNewStore();
 
 export const CancelBookingsDialog = (): JSX.Element => (
   <StoreProvider store={store}>
-    <Layout>
+    <Layout isAdmin={false}>
```
ikusteu, line 22: False should be the default

packages/ui/package.json
```diff
@@ -85,6 +85,7 @@
     "@eisbuk/shared": "workspace:*",
     "@eisbuk/svg": "workspace:*",
     "@eisbuk/translations": "workspace:*",
-    "luxon": "^2.0.5"
+    "luxon": "^2.0.5",
+    "react-router-dom": "~6.3.0"
```
ikusteu, line 89: This should be a peer dependency and should be externalized (I’ll explain)

ikusteu: Additionally, it should be v5.x as in the client for compatibility

packages/ui/src/Layout/AdminBar.tsx
```diff
@@ -0,0 +1,71 @@
...
+          })}
+      </div>
+    </div>
+  );
+};
+
+/** Get styles for top / botton row of the header */
+const getHeaderRowClasses = () => {
+  const baseClasses = [
+    "translate-x-80",
```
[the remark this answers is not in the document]
ikusteu: Especially explicit like ‘80’ or ‘2’

packages/ui/src/Layout/AdminBar.tsx
```diff
@@ -0,0 +1,71 @@
...
+    "translate-y-2",
+    "flex",
+    "h-10",
+    "text-white",
+    "hidden",
+    "md:flex",
```
ikusteu, line 52: I would also add md:justify-end

ikusteu: And have the links be on the right rather than awkward middle (translated to the right by 240px)

packages/ui/src/Layout/AdminBar.tsx
```diff
@@ -0,0 +1,71 @@
...
+    "items-center",
+  ];
+
+  return [...baseClasses].join(" ");
+};
+
+const getRoundedClass = (index: number, length: number) => {
```
ikusteu, line 59: Instead of writing JS logic to determine first/last, we can only use css and apply border radius on a parent element. That way the rounded edges will automatically obscure the edge buttons, plus it’s achieved using only CSS

packages/ui/src/Layout/Layout.stories.tsx
```diff
@@ -53,11 +54,17 @@ const user = {
     "https://images.unsplash.com/photo-1472099645785-5658abf4ff4e?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=facearea&facepad=2&w=256&h=256&q=80",
 };
 export const Default = (): JSX.Element => (
-  <Layout {...{ user }}>{dummyContent}</Layout>
+  <BrowserRouter>
```
ikusteu, line 57: It would be cleaner to add react router add on to storybook (I’ll take care of that)

packages/ui/src/Layout/Layout.tsx
```diff
@@ -1,15 +1,28 @@
...
   Notifications?: React.FC<{ className?: string }>;
   children?: React.ReactNode[] | React.ReactNode;
+  isAdmin: boolean;
+}
+interface Item {
+  Icon: string;
+  label: string;
+  slug: string;
+}
+enum PrivateRoutes {
```
ikusteu, line 22: This shouldn’t be hardcoded and is an app context (should be dynamically passed to the layout component from inside the app)

packages/ui/src/Layout/Layout.tsx
```diff
@@ -19,9 +32,34 @@ const Layout: React.FC<LayoutProps> = ({
   Notifications,
   children,
 }) => {
+  const { t } = useTranslation();
+
+  const adminsLink: Item[] = [
```
ikusteu, line 37: The comment about hardcoding the values instead of passing them to layout applies here as well

packages/ui/src/Layout/Layout.tsx
```diff
@@ -19,9 +32,34 @@ const Layout: React.FC<LayoutProps> = ({
...
+      Icon: People,
+      slug: PrivateRoutes.Athletes,
+    },
+  ];
+
   return (
     <div className="absolute top-0 right-0 bottom-0 left-0 flex flex-col overflow-hidden">
       <header className="bg-gray-800">
+        <div className="content-container">
+          <div className={getHeaderRowClasses("top")}>
```
ikusteu, line 59: This div can be part of the component (and recieve the classname through props)

packages/ui/src/Layout/AdminBar.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const AdminBar: React.FC<AdminBarProps> = ({ isAdmin, adminLinks }) => {
+  return (
+    <div>
+      <MobileHamburgerMenu adminLinks={adminLinks} />
+      <div className={getHeaderRowClasses()}>
+        {isAdmin &&
+          adminLinks.map(({ Icon, label, slug }, i) => {
+            return (
+              <div
```
ikusteu, line 23: Using a ‘Button’ component here would solve a couple of problems (hover:cursor-pointer, and avoid the div soup, plus it has a bunch of predetermined styles). Icon can be passed as ‘inputAdornment’.

ikusteu: Additionally Link should encapsulate the button (not just the label)

ikusteu (changes requested):
Go through the comments and apply fixes. Should be good to go.

ikusteu (approved):

PR #615, opened 2022-08-02
Rehaul/birthday menu admin layout

### LINKED ISSUE #608: Rehaul: Reinstate the Birthday menu in admin layout
## Rehaul: Reinstate the Birthday menu in admin layout

**Description:** While rehauling the admin app, we've omitted some stuff which had existed in the old app. One of those is the birthday menu. We should reinstate it: make it show next to the admin links bar. I would approach this by extending the Layout component (in `ui` package) so that it can render `additionalAdminContent` next to the admin links and then, in the `client` package, pass the birthday menu (as is) to the layout where needed.

### Requirements:

- [x] Extend the Layout component with `additionalAdminContent`:
  - it should be passed to `AdminBar` as `additionalContent` and rendered inline with the rest of the admin bar (after the admin bar)
  - it should be showcased in stories: a single additional `Button` will do (make sure to create a new story for that)
- [x] Show `BirthdayMenu` in the client app:
  - in the client app, pass the `BirthdayMenu` component as is (it should be rehauled, but that's a task for another day) to the Layout component on each page
  - meybe replace the existing SVG icon with a new SVG (can be exported from Librocco [icons](https://www.figma.com/file/yjspFXBKBIFJlC3wFv1FJL/The-Brocco?node-id=13%3A2602))

### LINKED ISSUE #616: Fix/birthdays-displayed-using-current-date
The birthdays are filtered/ordered using `currentDate` as a reference point. This is not desired behaviour, where birthdays should be shown with respect to today's date, while `currentDate` is used to navigate through the calendar, slots, etc.

### Desired behaviour:
Birthdays should alway be show with respect to "today", regardless of `currentDate` in store.

### Steps to reproduce:
To observe the faulty behaviour, try adding a couple of customers. Set one's birthday to yesterday, one's to today and one's to tomorrow (for good measure). At that point you want to switch current date with daily jump. For that you can go to `attendance` and scroll through the dates (using `CalendarNav`). You'll observe that the first birthday shown changes when the date in `CalendarNav` changes.

fadwamahmoud:
Fixes #608 
Fixes #616


packages/client/src/components/atoms/BirthdayMenu/BirthdayMenu.tsx
```diff
@@ -73,9 +73,12 @@ const BirthdayMenu: React.FC<BirthdayMenuProps> = ({ customers }) => {
         badgeContent={getTodaysBirthdays}
         data-testid={__birthdayMenu__}
       >
-        <IconButton onClick={handleBirthdaysClick} size="large">
+        <Button
+          onClick={handleBirthdaysClick}
+          className="h-8 w-8 !p-0.5 ml-2  hover:bg-black/10"
```
ikusteu, line 78: Loving the styling more/less. However, the user avatar below is 44px wide. It would be good to reflect that with the birthday menu button as items on the screen seem aligned that way. Secondly, when there's a badge, the button is almost not clickable.
Therefore, with a little playing around, I found this styling works nicely:
- width (and height) the same as user avatar: 44px
- padding: 8px (for the button to not appear too large and leave room for background to be shown on hover)
- leave the `ml-2` (works quite nicely)
- make the background white, instead of black on hover, as the background (of the header) is dark, opacity `10` works in that sense also (`hover:bg-white/10`)

ikusteu: Also, the `Badge` component needs `select-none` and `cursor-normal` classes (otherwise it shows a weird text select cursor over that tiny badge)

ikusteu (changes requested):
Some styling improvements are necessary (as I've described). There is also a larger issue, which has nothing to do with this PR and I'll open a different ticket for that.

PR #620, opened 2022-08-04
Rehaul/add to calendar

fadwamahmoud:
Reinstating Add to Calendar button back into the UI


packages/client/src/pages/customer_area/index.tsx
```diff
@@ -35,6 +41,10 @@ const CustomerArea: React.FC = () => {
   useSecretKey();
 
   const isAdmin = useSelector(getIsAdmin);
+  const date = useSelector(getCalendarDay);
+  const slotsByDay = useSelector(getSlotsForCustomer);
+
+  const bookedSlotsByMonth = useSelector(getBookedSlotsByMonth(date.month));
```
ikusteu, line 47: Correct me if I'm wrong, but I think you can use `getBookingsForCaledar` -> it'a an array of booked slots (full `SlotInterface` with booked `interval`), which would allow you to loop through those and create an ICalendar object. What this would achieve is:
- you wouldn't need to loop through bookings for monts and assign slots to each booking (so that you can read the `type`, as you would already have that data in the array you're looping over)
- it would reduce code "pollution" as you can get this data directly from the `AddToCalendar` instead of loading it to the page component, passing to the `AddToCalendar` component, etc.
- not tom mention it's slightly more efficient than the whole bunch iterations currently implemented and `getBookingsForMonth` can be completely deleted again

packages/client/src/store/selectors/bookings/slots.ts
```diff
@@ -176,3 +178,19 @@ export const getBookingsForCalendar = (
...
+export const getBookedSlotsByMonth =
+  (month: number) =>
+  (state: LocalStore): Record<string, CustomerBookingEntry> =>
+    Object.entries(state.firestore.data?.bookedSlots || {}).reduce(
+      (acc, [slotId, CustomerBookingEntry]) =>
+        DateTime.fromISO(CustomerBookingEntry.date).month === month
+          ? { ...acc, [slotId]: CustomerBookingEntry }
+          : acc,
+      {}
+    );
```
ikusteu, line 196: None of the changes to this file are necessary if you can apply the update to calendar creation logic I've described above

ikusteu (commented):
Added a couple of comments:
- visual stuff is a personal opinion
- the implementation stuff (especially) the passing around of excess data (bookings by month) should be corrected as to not introduce unnecessary entropy to the code

packages/client/src/pages/customer_area/index.tsx
fadwamahmoud: Yeah, definitely agreed on using `getBookingsForCalendar`. However, I think `bookedSlots` has to be loaded in the page component because `AddToCalendar` is conditionally rendered accordingly @ikusteu

ikusteu: You can just do: ‘return null’ if bookings are empty.

ikusteu (approved):

PR #622, opened 2022-08-06
Feature/textarea editable

### LINKED ISSUE #613: Feature: Text Area
Component: Textarea

**Description:**
A basic text area input field which allows users to enter text. 
Comes with an optional Label, Help Text, Placeholder.
An additional Actions Panel is defined to provide:
1. Supporting functionality - e.g Additional Help via "?", attachments, emoji keyboard...
2. Main "Confirm" | "Cancel" functionality to explicitly toggle edit / display of the text area content.
This is a presentational component - input state will be handled with Formik and passed to the component via the FormikField props.

Related to #505 

**Mocks:** [Textarea](https://www.figma.com/file/3ndhFlpGT36awa7RyOR5yK/Eisbuk-(EDIT)?node-id=2116%3A28902)

## Requirements
**Props should extend React.HTMLAttributes<HTMLTextAreaElement>:**
1. formikField - required: Formik.FieldProps (consists of Input, Meta and Form props)
3. label - optional: string
4. helpText - optional: string;
5. placeholder - optional: string
6. disabled - boolean = false;
7. showActions - boolean = false;
8. actions - optional: object: 
    1. supporting: array `{ Icon, onClick }[]`
    2. main: object 
        1. onConfirm: optional: callback function for _save_ main action
        2. onCancel: optional callback function for _cancel_ main action 

**Implementation**
1. The "disabled" state of the component represents its edit/display presentation. 
2. Actions panel should only be rendered if showActions = true & disabled != true
3. The onConfirm & onCancel "main action" icons should only be shown if respective handlers are passed via props. They can be used to explicitly toggle the disabled / display state of the field, as well as trigger side effects. Their action should only be enabled if there is change in the inputs value.
4. The "?" icon on the left of the mock is an example a "supporting action". If passed, the array of icons should be rendered in a container here.

**Styles**
1. Hover & Focus styles are detailed in [Textarea input](https://www.figma.com/file/3ndhFlpGT36awa7RyOR5yK/Eisbuk-(EDIT)?node-id=2042%3A28679) mock
2. CSS `resize: vertical` should be applied to textarea.

**Stories**
1.  Default - with label & helpText
3. w/ Actions Panel - showing main actions icons (cross & tick)
4. Display

ikusteu:
Fixes #613 
I used a slightly different approach than the one described in the spec. On initial pass I stuck closely to the spec, but have realised that such approach wouldn't work all that nicely in production (inside the app). Instead I've created a simple, but versatile, `TextareaEditable` component solely displaying text and switching to textarea when editing as well as integrating with Formik.
I'm continuing to pursue the `IntervalCardWithNotes` work and trying to reuse as much of the existing work as possible.

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -699,6 +699,7 @@ importers:
       eslint-plugin-react: ^7.29.4
       eslint-plugin-react-hooks: ^4.4.0
       eslint-plugin-unused-imports: ^2.0.0
+      formik: ^2.1.5
       jest: ^27.5.1
       luxon: ^2.0.5
       postcss: ~8.4.13
@@ -748,6 +749,7 @@ importers:
       eslint-plugin-react: 7.30.1_eslint@8.19.0
       eslint-plugin-react-hooks: 4.6.0_eslint@8.19.0
       eslint-plugin-unused-imports: 2.0.0_c296cfcf20d280cfb2cd3afc88aefdcf
+      formik: 2.2.9_react@16.14.0
       jest: 27.5.1
       postcss: 8.4.14
       react: 16.14.0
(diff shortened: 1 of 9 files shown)


delasous (approved):
Looks great! Love the way you use `React.createElement` with `displayedText`, and the test file is 👌🏼👌🏼👌🏼

ikusteu:
> Looks great! Love the way you use `React.createElement` with `displayedText`, and the test file is 👌🏼👌🏼👌🏼

🙏

PR #628, opened 2022-08-10
Rehaul/add to calendar review

ikusteu:


packages/client/src/components/atoms/AddToCalendar/AddToCalendar.tsx
```diff
@@ -93,27 +97,28 @@ const AddToCalendar: React.FC = () => {
...
+  // Don't render the component if there are no booked slots to save to calendar
+  if (!bookedSlots.length) return null;
+
+  return (
     <>
       <Button
-        className=" text-white bg-cyan-700 m-auto justify-center
-        position-relative
-         active:bg-cyan-800 md:w-auto  invisible md:visible "
+        className="hidden bg-cyan-700 text-white md:block active:bg-cyan-800"
```
ikusteu, line 106: 'hidden' is `display: none`, while 'invisible' is `display: hidden` (or something like that). The difference:

https://www.edureka.co/community/62212/explain-difference-between-visibility-hidden-display-none#:~:text=display%3Anone%20means%20that%20the,t%20seen%20on%20the%20page.

You can see the effect in dev tools (on small screen) of the original branch: the button isn't shown, but is still taking up space, while in this branch, the button is removed from the flow.

packages/ui/src/IconButton/IconButton.tsx
```diff
@@ -13,6 +15,7 @@ export enum IconButtonContentSize {
   Tight = "tight",
   Medium = "medium",
   Relaxed = "relaxed",
+  Loose = "loose",
```
ikusteu, line 18: You didn't have to do these updates (could have achieved the same result with only CSS classes passed to the `IconButton` component), but I took this opportunity to extend the `IconButton` component itself.

PR #635, opened 2022-08-31
Enhancement/sort attendance athletes

### LINKED ISSUE #633: In attendance view, sort athletes in a booking by booked slot
People who booked the same interval should be displayed next to each other.
Screenshot from 2022-08-22 11-59-59

ikusteu:
Fixes #633 
The customers are sorted by booked interval: longest first, then by start time
If the bookedIntervals are equal, the customers are sorted alphabetically: surname first, then the name

This still produces no results in `attendance_printable`, as there are mixed customers with and without booked slots. Since the `attendance_printable` displays `attendedInterval` as a fallback if the customer doesn't have a `bookedInterval`, we can apply the same logic to sorting and fall back to `attendedInterval` when comparing intervals. 
The current result is this:
- the customers who have booked are displayed at the bottom, without no visual divide, sorted as they should be
- the not-booked-but-attended customers are shown at the top (as `""` < `"10:00-11:00", or any other, defined, interval) and amongst themselves are sorted alphabetically
- the latter still have start and end time shown, but it's that of `attendedInterval`, while booked customers have the `bookedInterval` shown in the same place
- all in all `attendance_printable` looks completely unsorted

I have some ideas to tackle the issues, by bringing the sorting closer to the displaying component, but that would require some additional untangling and could be done as an enhancement.


ikusteu:
Included in #648

PR #638, opened 2022-09-01
Change customer category to an array

fadwamahmoud:


packages/firestore/firestore.rules
```diff
@@ -132,13 +132,14 @@ service cloud.firestore {
...
             bookingData.interval in bookedSlot.data.intervals     
+            bookedSlot.
           );
         }
         function checkBookedSlotCategory(slotId) {
           let customer = get(/databases/$(database)/documents/organizations/$(organization)/bookings/$(secretKey));
           let bookedSlot = get(/databases/$(database)/documents/organizations/$(organization)/slots/$(slotId));
           
-          return customer.data.category in bookedSlot.data.categories
+          return bookedSlot.data.categories.hasOnly(customer.data.category)
```
ikusteu, line 142: Ok, so here I would do `customer.data.category.hasAny(bookedSlot.data.categories)`

It works both ways, but this way it's clear that you're checking customer, not a slot

packages/firestore/firestore.rules
```diff
@@ -172,7 +173,7 @@ service cloud.firestore {
           checkOptionalValidDate(request.resource.data, "certificateExpiration") &&
           checkOptionalValidDate(request.resource.data, "covidCertificateReleaseDate") &&
           // Check category (allows updates even if "adults")
-          request.resource.data.category in getAllCategories() &&
+          request.resource.data.category.hasAny(getAllCategories()) &&
```
ikusteu, line 176: Here you should use `hasOnly`. A case we want to prevent: `customer.data.category = ["course", "non-existing-category"]`

packages/client/src/__tests__/customerDataTriggers.test.ts
```diff
@@ -94,20 +94,25 @@ describe("Customer triggers", () => {
...
       // check `bookings` entry updates
       await waitForCondition({
         documentPath: getBookingsDocPath(organization, secretKey),
         condition: (data) =>
           Boolean(
-            data && data.category === updatedCategory && data.deleted === true
+            data &&
+              data.category.every((cat: Category) =>
+                updatedCategory.includes(cat)
+              ) &&
```
ikusteu, line 114: Here, if you want to test that the updated bookings doc includes all `updatedCategories`, you wouldn't loop over the bookings doc's categories and check that they're included in `updatedCategories`, but rather the other way around: you want to assert that each of the `updatedCategories` found its way to `data.category`.

fadwamahmoud:
> Nice! I tested the migration and it worked perfectly. When trying to edit an (or add a new) athlete, the form is not showing labels for categories: Screenshot from 2022-09-05 16-42-54 Also, if I check more than one, they are concatenated with no separator. They should be separated by comma-space (`, `): Screenshot from 2022-09-05 16-42-42

Apparently, labels were actually showing up for me because the enum keys are in English...
But yeah, fixed them both!

packages/client/src/__tests__/migrations.test.ts
```diff
@@ -245,4 +245,59 @@ describe("Migrations", () => {
       ).rejects.toThrow(HTTPSErrors.Unauth);
     });
   });
+
+  describe("migrateCategoriesToArray", () => {
```
ikusteu, line 249: Yeah...this is what I was talking about, `migrateCustomersToMultipleCategories` or `migrateCustomersToCategoriesArray` would make more sense. Or namespaced version (as you've suggested), something like this: `customersToPluralCategories`

packages/client/src/__tests__/migrations.test.ts
```diff
@@ -245,4 +245,59 @@ describe("Migrations", () => {
...
+    testWithEmulator("should change category to array", async () => {
```
ikusteu, line 250: Aand this should then be more verbose (I'm using this as an example, don't dwell too much on it as it will probably be removed as soon as all of the data is migrated)

packages/client/src/__tests__/migrations.test.ts
```diff
@@ -245,4 +245,59 @@ describe("Migrations", () => {
...
+        );
+        await customerRef.set(customer);
+
+        await expect(
+          invokeFunction(CloudFunction.MigrateCategoriesToArrays)({
+            organization,
+          })
+        ).resolves.not.toThrow();
+      }
+    );
```
ikusteu, line 293: I think this error is in place. Maybe, in this case, I would test that the updates simply aren't applied to customers which have already been migrated (should be aplicable to all migration cases).

packages/client/src/components/atoms/CustomerCard/CustomerCard.tsx
```diff
@@ -70,10 +70,15 @@ const CustomerCard: React.FC<CustomerCardProps> = ({
...
-                  /-/g,
-                  " "
-                )
+              ? customer.category.map((cat, i) => {
+                  const separator =
+                    i === customer.category.length - 1 ? "" : ", ";
+
+                  return `${capitalizeFirst(t(CategoryLabel[cat])).replace(
+                    /-/g,
+                    " "
```
ikusteu, line 79: You can omit whitespace and separator by simply adding `categories.map(...).join(", ")` and that handles the first/middle/last element check for you.

packages/client/src/__testData__/customers.ts
```diff
@@ -12,7 +12,7 @@ export const saul: Customer = {
   birthday: "2001-01-01",
   covidCertificateReleaseDate: "2021-01-01",
   covidCertificateSuspended: true,
-  category: Category.Competitive,
+  category: [Category.Competitive],
```
silviot, line 15: I think we should change the `category` name to `categories`.
This probably involves supporting both names in type declarations.
After the migration is done  we can drop `category` from the code.

silviot:
> But yeah, fixed them both!

Confirmed! Both are fixed now. After today's discussion about the importance of naming I came to the conclusion we should really rename `category` (scalar) to `categories` (array). I understand it's a lot of work, but it's necessary to have code that doesn't confuse you (we have a `category` field in slots, and it's a scalar).

packages/client/src/components/customers/CustomerForm/CustomCheckbox.tsx
```diff
@@ -9,16 +9,23 @@ import makeStyles from "@mui/styles/makeStyles";
...
-const CustomCheckbox: React.FC<Props> = ({ name, label }) => {
+const CustomCheckbox: React.FC<Props> = ({
+  name,
+  label,
+  disabled = false,
+  value,
+}) => {
   const classes = useStyles();
-  const [field] = useField({ name, type: "checkbox" });
+  const [field] = useField({ name, type: "checkbox", value });
```
ikusteu, line 23: Is the value necessary here...my understanding is that it is stored in formik's state and retrieved + updated through `field` object.

ikusteu: I see...(checkboxes and radios always eluded me). Disregard this ☝️

packages/client/src/store/actions/__tests__/bookingOperations.test.ts
```diff
@@ -120,6 +120,7 @@ describe("Booking Notifications", () => {
         const bookedSlotsForCustomer = await getDocs(
           collection(db, getBookedSlotsPath(organization, secretKey))
         );
+        console.log({ bookedSlotsForCustomer });
```
ikusteu, line 123: I think this is a forgotten debug artefact

packages/client/src/store/selectors/bookings/__tests__/bookings.test.ts
```diff
@@ -43,7 +43,7 @@ const setupBookingsTest = ({
   date,
   slotsByDay,
 }: {
-  category: Category | DeprecatedCategory;
+  category: Category[] | DeprecatedCategory[];
```
ikusteu, line 46: To avoid confusion here, let's move this logic down to `updateCustomer` part 👇 
props: `category: Category | DeprecatedCategory`, and then, below when creating customer, `...getCusrtomerBase(saul), category: [category]`

packages/client/src/store/selectors/bookings/__tests__/bookings.test.ts
```diff
@@ -308,7 +308,7 @@ describe("Selectors ->", () => {
   describe("getMonthEmptyForBooking", () => {
     interface TestParams {
       name: string;
-      category: Category;
+      category: Category[];
```
ikusteu, line 311: Same goes here (using the category as is, singular)

packages/functions/src/migrations.ts
```diff
@@ -172,20 +172,57 @@ export const migrateCategoriesToExplicitMinors = functions
...
+
+      if (!Array.isArray(data.category)) {
+        functions.logger.info(`Converted customer: ${data.id}`);
+
+        batch.set(customer.ref, { category: [category] }, { merge: true });
       }
-      batch.set(customer.ref, { category }, { merge: true });
+      // else {
+      //   functions.logger.error(`Cannot convert customer ${data.id} category as it already is an array`)
+      // }
```
ikusteu, line 225: Seems left behind

ikusteu (dismissed):
LGTM, aside from some small stuff, which I've used to illustrate a point (no need to fix them for this unit of work).

Do fix the `setupBookings` -> singular category and all it entails

packages/client/src/__tests__/migrations.test.ts
fadwamahmoud: Yeah, that's exactly what I attempted to do at first but couldn't think of a way to achieve it. So, any tips would be appreciated.

silviot (approved):

PR #639, opened 2022-09-02
Fix/email sending bug 1

ikusteu:
A fix for email sending bug, where emails were passed around in the form:
```typescript
{
  from: string
  to: string
  message: {
    subject: string
    html: string
    attachments: Attachment[]
  }
}
```
to (more convenient) form:
```typescript
{
  from: string
  to: string
  subject: string
  html: string
  attachments: Attachment[]
}
```

I don't know if this was a `nodemailer` change, or a simple misconstruction on our part, but now it works.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -8,9 +8,7 @@
       "cd packages/functions && rushx build",
       "sed -i '/\"workspace:\\*/d\"' \"$RESOURCE_DIR/package.json\""
     ],
-    "postdeploy": [
-      "sed -i -e \\'s/  .dependencies.: {/  \"dependencies\": {\\n    \"@eisbuk\\/shared\": \"1.0.0\",/\\' \"$RESOURCE_DIR/package.json\""
-    ],
+    "postdeploy": ["git restore packages/functions/package.json"],
     "source": "packages/functions"
   },
   "hosting": [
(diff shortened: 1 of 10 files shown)


firebase.json
```diff
@@ -8,9 +8,7 @@
       "cd packages/functions && rushx build",
       "sed -i '/\"workspace:\\*/d\"' \"$RESOURCE_DIR/package.json\""
     ],
-    "postdeploy": [
-      "sed -i -e \\'s/  .dependencies.: {/  \"dependencies\": {\\n    \"@eisbuk\\/shared\": \"1.0.0\",/\\' \"$RESOURCE_DIR/package.json\""
-    ],
+    "postdeploy": ["git restore packages/functions/package.json"],
```
silviot, line 11: TIL about `git restore`! It's been around [for three years now](https://github.blog/2019-08-16-highlights-from-git-2-23/#experimental-alternatives-for-git-checkout), but I hadn't cought up yet. Thanks!

ikusteu: Np. I use `git restore --staged <staged-file-or-pattern>` to unstage the file or in this context to restore an unstaged file to the version of the file on the last commit. 

And I always mix it up with `git reset` (now only in parlance and in typing, in the past I would also mix up the functionalities)

ikusteu:
Included in #644

Issue #574, opened 2022-07-05
Delete checklist (client package)


ikusteu:
# Delete checklist (client package)

This is a checklist of all artifacts (components, hooks, functions) that can be deleted as being replaced by rehauled versions or have become obsolete.

I've imagined this ticket as being open for the duration of Admin incremental rehaul, rebased each time one of the dependency tickets has been closed and deleting artefacts as soon as they have been replaced.

## Components:

- [x] `atoms/BookingCard`
- [x] `atoms/BookingCardGroup`
- [x] `atoms/BookingsCountdown`
- [x] `global/ConfirmDialog` - depends on #569 
- [x] `layout/*` - TBD
- [x] `auth/*` - TBD
- [x] `users/EisbukAvatar` - depends on #570 
- [x] `customers/CustomerForm` - depends on #572 
- [x] `Notifier.tsx` - depends on #571 
- [x] `atoms/NotificationButton.tsx` - depends on #571 
- [x] `atoms/AttendanceCard/*` - depends on #573 
- [x] `atoms/CustomerCard/*` - depends on #572 
- [x] `atoms/CustomerList/*` - depends on #573 
- [x] `atoms/DateNavigation/*` - depends on #570 
- [x] `atoms/SlotsDaycontainer/*` - depedns on #568 
- ...more TBD

## Store:

- [x] `store/actions/appActions.ts`:
  - depends on #<notifications replace>
  - artifacts to delete:
    - enqueueNotification
    - closeSnackbar
    - removeSnackBar
    - showErrSnackbar
    - appReducer should be updated to exclude the aforementioned actions
    - getNotifications in `store/selectors/app.ts`

## Dependencies:
- [x] `notistack`:
  - depends on #571

PR #640, opened 2022-09-02
Remove .env file as it's unnecessary

ikusteu:
_I could have YOLO merged this, but I would like everybody to read this update:_

I could talk days on end how our deployment stages are currently a bit convoluted, but that's a problem for another ticket.

This removes the contents of `.env` file in `packages/client` as the file is used to provide (mostly dummy) secrets for development. The fallbacks, however, are provided by `vite.config.ts` for dev setting and the code provides fallbacks used for testing.

**edit:** this file isn't removed completely, but its contents are removed: all but `STORYBOOK_IS_STORYBOOK=true`, as that env variable is only picked up by storybook build and we use it to keep some things consistent (like the date) in storybook builds not to produce false positives in chromatic checks.

**This file (.env) should be populated manually by the secrets provided in 'secrets' slack channel if you wish to use the staging environment (and firebase project)**


ikusteu:
Requires more work to be green -> should be done altogether with #574

PR #647, opened 2022-09-03
Enhancement/booked slot disable delete

### LINKED ISSUE #631: Disable deletion for slots that have bookings
Deleting a slot with bookings will break the app. We should not let admins do it. The delete button should be grayed out when there are bookings for one slot.
The app will break when an athlete who booked the deleted slot tries to manage their bookings for that period.

ikusteu:
Fixes #631 
- Created `DeleteSlotDisabledDialog`
- Update of `ActionDialog` to render buttons conditionally with respect to confirm/cancel label being present
- Updated `SlotCard` and `SlotOperationButtons/DeleteButton` to accept `disableDelete` and if `true`, show `DeleteSlotDisabledDialog` modal instead of proceeding with slot deletion
- Painted `SlotOperations/DeleteButton` lighter gray to indicate it being disabled-ish
- Applied logic to differentiate booked slots in `slots` page (by cross referencing with `attendance` entries in store)

### DIFF
--- a/packages/client/src/components/atoms/SlotCard/SlotCard.tsx
+++ b/packages/client/src/components/atoms/SlotCard/SlotCard.tsx
@@ -35,6 +35,10 @@ export interface SlotCardProps extends SlotInterface {
    */
   enableEdit?: boolean;
   /**
+   * Disable slot deletion (we use this if there are booked intervals on a particular slot)
+   */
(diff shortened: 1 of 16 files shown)


silviot (approved):
LGTM. I updated the Italian translations.

PR #650, opened 2022-09-04
Enhancement/auth tailwind phone prefix

### LINKED ISSUE #634: Accept `00` as prefix in login form (currently only `+` is accepted)
This is displayed if you dare use old-style prefix:
Screenshot from 2022-08-22 12-25-30

ikusteu:
Fixes #634 
Doesn't really fix it, but provides a workaround where we use `TextInput` with country code prefixed (and added as a prefix to final string) when typing in the phone number.


ikusteu:
@silviot yes...the country dropdown needs more work in order to be implemented...for that I would first go over the `react-redux-firebase-auth` tidy-up/upgrades (which makes it easier). 
Furthermore, `TextInput` would need upgrade where the dropdown is an option and the adding of the prefix/dropdown value is handled under the hood (so that we get `onChange` with the full value, not only the value of the field, as the prefix then serves only as a presentation layer sugar). The workaround, for this case, would be to wrap it in `AuthTextField` which came to me just now and am happy to do it.
Also, I completely forgot to add it to the customer form (only the login form), so, on it.

As for the available implementation: you can either update the `publicOrgInfo` manually if you want to test it, or deploy the functions as the `createPublicOrgInfo` (should be called `updatePublicOrgInfo` in my opinion) is updated to copy over the `countryCode` to `publicOrgInfo` as well. 
The flow:
- user updates the `countryCode` using the UI
- the update is stored in `organization` document (not accessible to non-admins)
- the data trigger copies some values over to `publicOrgInfo` to be publicly accessible (this part requires function deployment)

So there you have it, if you want to test it out. As for me...I will make an effort to:
- [ ] add the prefix to customer form
- [ ] replace the prefix with the dropdown `AuthTextField`

The further enhancements, I think, would best be done on the `TextInput` directly, but I can provide a workaround in customer form as well.

PR #657, opened 2022-09-08
Feature/phone input

ikusteu:
Fixes #634 
- Extracts `DropdownAdornment` as a separate (`Dropdown`) component
- Creates a `CountryCodesDropdown` component (to select country dial code)
- Creates a `PhoneInput` component using the `CountryCodesDropdown` to prefix the phone in put
- Adds `defaultCountryCode` input in `admin_preferences` and adds it to `publicOrgInfo` (adding to existing data trigger)
- Applies the `PhoneInput` component to be used as phone version of `AuthTextInput` and in `admin_preferences` (when setting up the `defaultCountryCode`

Additionally, the `admin_preferences` page is a bit messy and the `PhoneInput` should be used in `CustomerForm`. I will attend to the latter, and the former (if time permits) to apply the, least painful, quick rehaul utilising the components from the `ui` package.

PR #659, opened 2022-09-10
Rehaul/old customer form refresh

ikusteu:
Includes #657  (so it should be merged after)

Polished up the old CustomerForm with the ui components (removing MUI) and applied some fixes/extensions to form elements in the ui package.


silviot (changes requested):
Nice! Looks great, apart from a couple of needed fixes.
The last field (card number) is triggering browser autocomplete for credit cards. An `autocomplete="off"` could solve this.
More importantly, category labels all point to the first category, so if you click on any category label the first checkbox will get triggered.

silviot:
Athlete currently can't see their data: https://eisbuk--rehaul-admin-preferences-refresh-4be8l1uj.web.app/customer_area/c5a16cc6-26c2-4ed8-9749-1641ae418181

Also, logged in as admin I was not able to save this form: when I click "save" nothing happens.

delasous (approved):

silviot (approved):

PR #661, opened 2022-09-12
AttendedSlots subcollection

fadwamahmoud:


packages/functions/src/dataTriggers.ts
```diff
@@ -269,7 +269,82 @@ export const createAttendanceForBooking = functions
...
+  .onWrite(async (change, context) => {
+    const { organization, bookingId } = context.params as Record<
+      string,
+      string
+    >;
+
+    const db = admin.firestore();
+
+    const isCreate = !change.before.exists;
+    // const isDelete = !change.after.exists;
```
ikusteu, line 317: Ok...so here, it would be useful to get `isCreate` and `isUpdate`. If it's create, you would update each customer's attended slots, and on update, only the updated (that can be achieved by comparing `before` and `after` data)

packages/functions/src/dataTriggers.ts
```diff
@@ -269,7 +269,82 @@ export const createAttendanceForBooking = functions
...
+
+    if (isCreate) {
```
ikusteu, line 319: Ok, but what if the attendance is updated?

packages/functions/src/dataTriggers.ts
```diff
@@ -269,7 +269,82 @@ export const createAttendanceForBooking = functions
...
+      const attendanceData = change.after.data() as SlotAttendnace;
+
+      const customerId = Object.keys(attendanceData.attendances)[0];
```
ikusteu, line 322: This would update only the attended slots for the first customer attended (or booked) that day, as the attendance contains the following record:
```typescript
{
  [slotId: string]: {
    attendances: {
      [customerId: string]: {
        bookedInterval: string | null
        attendedInterval: string | null
      }
    }
  }
}
```

ikusteu: You'd want to loop over all of the customers and update their attendances (but only if needed)

packages/functions/src/dataTriggers.ts
```diff
@@ -269,7 +269,82 @@ export const createAttendanceForBooking = functions
...
+//   /**
+//    * Booked interval for particular slot
+//    */
+//   interval: string;
+
+// }
+export const createAttendedSlotOnAttendance = functions
+  .region(__functionsZone__)
+  .firestore.document(
+    `${Collection.Organizations}/{organization}/${OrgSubCollection.Attendance}/{bookingId}/`
```
ikusteu, line 306: Let's call the document id `slotId`, as it's used to key the attendance entries, but also to group (and point back to) the given slot.

packages/functions/src/dataTriggers.ts
```diff
@@ -269,7 +269,84 @@ export const createAttendanceForBooking = functions
...
 
+    // const isCreate = !change.before.exists;
+    const isUpdate = change.after.exists;
+    // const isDelete = !change.after.exists;
+
+    if (isUpdate) {
+      const attendanceData = change.after.data() as SlotAttendnace;
+
+      Object.keys(attendanceData.attendances).forEach(async (customerId) => {
+        const attendances = attendanceData.attendances[customerId];
```
ikusteu, line 324: It is better to replace `.forEach` with `.map` and return 0 or null, or whatever from each map. Then to `Promise.all` on the result, e.g.:
```typescript
const updates = someArray.map(async () => {
  await someAsyncTask()
  return null
}

await Promise.all(updates)
```

ikusteu: This returns an array of promises which all resolve to `null`, but `Promise.all` can be utilised to await all of the async tasks

ikusteu: Or better yet:
```typescript
const updates = someValues.map(() => new Promise<void>(async (resolve, reject) => {
  await someAsyncTask()
  resolve()
}))
```

This way you're explicit about returning a promise from each map function and are not relying on syntactic sugar which might, or might not behave as you assume, so I prefer that approach.

_notice no `async` in the map function and `async` being introduced only in the promise callback_

PR #664, opened 2022-09-14
Display unbooked attended lessons in athlete calendar view

fadwamahmoud:


packages/functions/src/dataTriggers.ts
```diff
@@ -338,3 +338,100 @@ export const createPublicOrgInfo = functions
...
+      if (hasPreviousBooking || isBooking) return { [id]: "noUpdate" };
+
+      return currentAttendanceData.attendances[id]
+        ? previousAttendanceData.attendances[id]
+          ? currentAttendances[id].attendedInterval ===
+            previousAttendances[id].attendedInterval
+            ? { [id]: "noUpdate" }
+            : { [id]: "update" }
+          : { [id]: "update" }
+        : { [id]: "delete" };
```
delasous, line 392: I find nested ternary's like this very difficult to decipher: I've been reading this for a few minutes and it still isn't very clear to me what going in and causes what to come out. I would suggest that if it goes deeper than 2 "?"'s, try to compose the end result from individual conditional statements.

delasous (approved):

Issue #656, opened 2022-09-08
Allow athletes to self register


silviot:
Athletes should be able to create an entry for themself without any action required by admins.
The flow should be:
* Athletes are provided with a URL and an organization-wide weak secret (a passphrase that can be changed in the organization settings)
* Athletes log into the system (actually they create a user) with the existing form
* If there is no athlete entry for the currently logged in user, we present a form to fill out. Email or phone will be pre-filled with the value used for login.
* Upon succesful form submission a new entry will be created in the `customers` collection, and the trigger will in turn assign it a secret key and create an entry in the `bookings` collection with the secret key. The user will be redirected to their booking management page at this point.

The newly created user will have all categories associated with them, so the system should allow them to book any slot (until an admin comes in and assigns categories to restrict what they can book).

## Task breakdown

The initial step is already developed. If a user doesn't exist, they can register and a new auth user will be created. There are still a couple of things we should do update in that regard:

- [x] Differentiate sign in and register buttons on inital auth screen
- [x] Don't require `name and surname` field, or make it two fields: `name` and `surname` (anybody's input/preference appreciated here)

After the user is registered, a modal/slide-over/route for registration should be open (we can probably reuse the UI for CustomerProfileForm) at which point, after all of the required fields are filled, the user will get created in firestore and await admins "approval" (assigning a category) at which point we should probably show some message stating: you haven't been assigned to any category, contact your admin, etc.

### UI

- [x] Reuse the `CustomerProfileForm`:
  - we will need to decouple the logic of the "Edit" button/state (should probably lift it out) as it's unnecessary for self registration logic
  - my suggestion would be to create a route for self registration, which would present the `CustomerProfileForm`, without the `isEditing` toggle (it would always be `true`) as that seems to be the quickest solution
- [x] Add a field for orgnization reigster secret
- [x] Display a message (to the user if they've been reigstered succesfully, but are not yet assigned a category)

### Cloud function (https):

- [x] Validates register secret
- [x] Validates user fields and creates a new user in firestore (if the fields are correct)
- [x] Sends an email to the admin that there's a new user in queue for approval (category assignment)

ikusteu:
Hey @silviot, this is mostly implemented, but I see a small discrepancy between the implementation and the requirements mentioned here:

> The newly created user will have all categories associated with them, so the system should allow them to book any slot (until an admin comes in and assigns categories to restrict what they can book).

[...]

PR #672, opened 2022-09-20
Fix/populate customer form

### LINKED ISSUE #667: Customer form is not populated for non-admin users
The customer form does not load customer data when a non-admin user navigates to it.
For example: https://eisbuk--rehaul-admin-preferences-refresh-4be8l1uj.web.app/customer_area/c5a16cc6-26c2-4ed8-9749-1641ae418181

- [ ] The [data trigger](https://github.com/eisbuk/EisBuk/blob/master/packages/functions/src/dataTriggers.ts#L34) that copies over data to the `bookings` collection should be updated to [include more details](https://github.com/eisbuk/EisBuk/blob/390205dd6438e98fcd1bf830eb63ae1ffb75c17d/packages/shared/src/utils/customers.ts#L9), i.e. all the ones users should be able to see in the form
- [ ] The form should be driven by the data mentioned above
- [ ] On form submit a (to be implemented) cloud function should be invoked
- [ ] The new cloud function should use a secret key provided in its invocation to mutate the original item in the `customers` collection

fadwamahmoud:
Fixes #667 
Solution is not ideal but is the quickest.
Hopefully when we work on #656 we could implement a better way to access data based on user roles.

### DIFF
--- a/packages/client/src/__tests__/firestoreRules.test.ts
+++ b/packages/client/src/__tests__/firestoreRules.test.ts
@@ -519,28 +519,31 @@ describe("Firestore rules", () => {
   });
 
   describe("Customers rules", () => {
-    testWithEmulator("should only allow admin access", async () => {
-      const { db, organization } = await getTestEnv({
-        auth: false,
-        setup: (db, { organization }) =>
-          setDoc(doc(db, getCustomerDocPath(organization, saul.id)), saul),
-      });
-      // check read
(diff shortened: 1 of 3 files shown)


silviot:
@fadwamahmoud I updated https://github.com/eisbuk/EisBuk/issues/667 with a description of the plan we discussed earlier.

PR #673, opened 2022-09-20
Feat/attendance report page

### LINKED ISSUE #632: Let admins view past attendance info, highlighting differences between what was booked and what was attended
It should look like this:

Each pair of rows is about an athlete. The first shows booked and attended times.
The second shows departures from that. A booked hour that was not attended will result in a `-1` for that athlete on that day. An unbooked hour that was attended will result in a `1`.

The implementation should first be an html table.
We could then use [sheetjs](https://docs.sheetjs.com/docs#export-an-html-table-to-excel-xlsx) for the heavy lifting around the Excel format.
Taking care of just the HTML table and do the excel part as a later task is also perfectly acceptable, in case sheetjs proves difficult to use.

delasous:
Fixes #632


silviot:
It' really nice! 
There are a couple of problems that need fixing before this can go in:
- There should be no fractional values in the table, except halves. 
  
- The layout should have a single scrollbar. Currently there's an extra empty space at the bottom, and two scrollbars. More apparent on desktop than on mobile.

PR #677, opened 2022-09-26
Fix/customer edit form

fadwamahmoud:


packages/client/src/store/actions/bookingOperations.ts
```diff
@@ -130,6 +134,50 @@ export const updateBookingNotes: UpdateBooking<{ bookingNotes: string }> =
...
+    secretKey: Customer["secretKey"];
+    customer: Customer;
+  }): FirestoreThunk;
+} =
+  ({ secretKey, customer }) =>
+  async (dispatch, getState) => {
+    const organization = getOrganization();
+
+    try {
+      const db = getFirestore();
```
silviot, line 154: When the form is submitted we want to invoke a cloud function. See third point "[On form submit a (to be implemented) cloud function should be invoked](https://github.com/eisbuk/EisBuk/issues/667)" on #667.

The `updateBookingCustomer` action should be changed to invoke a (new, to be written) cloud function, instead of mutating firestore entries from the client.

packages/functions/src/dataTriggers.ts
```diff
@@ -440,3 +440,54 @@ export const createAttendedSlotOnAttendance = functions
...
+// check if update change.after.exists && contains customer data not booked slots
+// update customer doc in customers collection
+
+/**
+ * Data trigger used to update attendance entries for slot whenever customer books a certain slot + interval.
+ *
+ * - listens to `organizations/{organization}/bookings/{secretKey}/bookedSlots/{slotId}`
+ * - writes to `organizations/{organization}/attendnace/{slotId}` - updates entry for `attendances[customerId]` leaving the rest of the doc unchanged
+ */
+export const updateCustomerWithDataFromBookings = functions
```
silviot, line 454: This trigger should not be needed. Athletes editing a form will hit a cloud function on submit. The cloud function will check that they provided a valid secret key, and if they did it will update the `customers` collection.
The (poorly named, my bad!) `addIdAndSecretKey` method will take care of copying the new data to the `bookings` collection.
This is to make sure we have one authoritative, primary place to hold the data (the `customers` collection) and the other spot (the `bookings` collection) is treated as secondary, and can only get its values from the primary source.

I also expected this to create a "trigger loop": `updateCustomerWithDataFromBookings` will trigger when there's an update to `bookings`, and it will update `customers` accordingly. This in turn will cause the `addIdAndSecretKey` trigger to run; it will update the `bookings` collection accordingly. This in turn will trigger `updateCustomerWithDataFromBookings`, rinse and repeat...
The above is not happening and I'm not sure why. Still, I firmly believe in the necesssity to have one place as primary and another one as secondary, and only mutate the primary one via client code. The secondary one will be updated only by triggers.

silviot (changes requested):
I think there's been a misunderstanding regarding the preferred way to update customer data. I pushed a change (removing update permission for users) that is making tests red. Changing the form submission to use a cloud function (as opposed to hitting firebase directly) will fix it.

packages/functions/src/dataTriggers.ts
silviot: Also, I believe the function docs were not updated after copy pasting. No need to fix them, since this trigger should go.

silviot (approved):

PR #680, opened 2022-10-17
Enhancement/modal component update

ikusteu:
Allows for a modal component to get updated by outside change. Instead of dispatching `openModal` we're using `useModal` hook, passing the props to the hook (the updates to props get automatically propagated to the modal component), and firing `open` to open the modal, or simply firing `openWithProps` to set the modal component props directly (for modals which won't get updated by outside change, e.g. confirmation dialogs, etc.)

**Example 1 (automatic update)**
```typescript
const useModal = createModal("modal-component-name")

// In parent component
const { open } = useModal(componentProps)

// open modal on some action/state update
open()
```

**Example 2 (explicit props set)**
```typescript
const useModal = createModal("modal-component-name")

// In parent component
const { openWithProps } = useModal()

// open the modal with props explicitly set
openWithProps(modalProps)
```

### DIFF
--- a/packages/client/src/__testUtils__/testHooksWithRedux.tsx
+++ b/packages/client/src/__testUtils__/testHooksWithRedux.tsx
@@ -1,0 +1,64 @@
+import React, { useEffect } from "react";
+import { Provider } from "react-redux";
+import { Store } from "redux";
+import { render } from "@testing-library/react";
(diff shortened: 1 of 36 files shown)


silviot (approved):
LGTM

PR #684, opened 2022-11-17
Feature/email copy test

fadwamahmoud:


packages/client/src/__tests__/cloudFunctions.test.ts
```diff
@@ -32,7 +34,46 @@ describe("Cloud functions", () => {
...
-  describe("sendMail", () => {
+  describe("test email sending", () => {
+    // Dummy data for error testing
+    const to = "saul@gmail.com";
+    const subject = "Subject";
+    const html = "html";
+    const smtpServer = createJestSMTPServer({ port: 5000, host: "localhost" });
+
+    afterAll(() => {
+      smtpServer.close();
```
fadwamahmoud, line 45: If you comment out this line, you'd get more logs including a connection from your machine (this is why I suspect the connection closes too soon) but then you'd have to close the port manually after each run. @silviot

ikusteu:
I've pushed fixes...while I applaud the upgrading of `jest-smtp`, I've made a very simple fix (in usage, not implementation) for our use case). We don't need any additional updates so, while the library could use some updates in both matchers and recording of sent emails, I think that if we wanted to participate, we should do so outside the scope of eisbuk project (like @delasous has planned for `svelte-headlessui`) and I would much rather remove the code for smtp jest copied over to our codebase.

@fadwamahmoud @silviot what are your thoughts on this?

packages/client/src/__testSetup__/node.ts
```diff
@@ -1,21 +1,26 @@
...
-  (doLogin?: boolean): Promise<{
+  (doLogin?: boolean, additionalSetup?: Partial<OrganizationData>): Promise<{
     organization: string;
     email: string;
     pass: string;
   }>;
 }
 
-export const setUpOrganization: SetUpOrganization = async (doLogin = true) => {
+const smtpPort = 5000;
```
ikusteu, line 18: @silviot I've moved this here as we won't need smtp setup while testing client code as tests testing cloud functions (in charge of sending emails) are always ran in node environment.

PR #690, opened 2022-12-02
Implement email templates

### LINKED ISSUE #681: Allow admins to specify templates for emails that are sent out: 8 pts
Currently there's an email template in the admin preferences, but it has no effect.
We should change it with a version that allows templating and preview.

Admins should see the available placeholders (`${first-name}`, `${lsat-name}` etc) and see an example render of the template. Placeholders could be shown as buttons that, when clicked, fill in their text in the current cursor position.

## Tasks

- [ ] Solve this in the UI:
  - We can use the organisation preferences page
  - Replace `emailTemplate` field with `emailTemplates`
  - The saved templates could be shown as truncated text (cut off with "..." at the end)
  - The field should be clickable so that the modal with email templates gets opened
  - The modal could have a dropdown selection of all possible templates
  - The template currently edited (in modal) should render an input field for the template itself, short info message and a preview with dynamic fields replaced by some dummy values
  - The function used to process the template and show a preview can be reused later for constructing of an email message
  - The templates should be initialised with default values and submission of an empty form shouldn't be acceptable (both client side and firesore.rules)

- [ ]  Solve this in the data model:
  - The email templates should be stored as a record of template name and template value
  - The templates should be initialised with default values

- [ ] Consume these templates when sending out emails (should we move the template interpolation to a cloud function and send only the inputs: name, surname, etc) or interpolate on the client:
    - I would go with cloud function as some automatic emails might be triggered by an unauth user, so it makes sense to keep the templates to cloud function access

## Updated tasks

_I've kept the tasks above for reference_

- [x] Set up email payload types (with email template being retrieved using email type):
```typescript
// packages/shared/src/enums/email.ts
export enum EmailType {
  SendBookingsLink = "send-bookings-link",
  SendCalendarFile = "send-calendar-file",
  SendExtendedBookingLink = "send-extended-booking-link",
}

// packages/shared/src/types/email.ts
interface EmailTypePayload {
  [EmailType.SendBookingsLink]: {
    type: EmailType.SendBookingsLink;
    organization: string;
    customer: {
      name: string;
      surname: string;
      secretKey: string;
    };
  };
  [EmailType.SendCalendarFile]: {
    type: EmailType.SendCalendarFile;
    organization: string;
    customer: {
      name: string;
      surname: string;
      secretKey: string;
    };
  };
  [EmailType.SendExtendedBookingLink]: {
    type: EmailType.SendExtendedBookingLink;
    organization: string;
    bookingsMonth: string;
    extendedBookingsDate: string;
    customer: {
      name: string;
      surname: string;
    };
  };
}

export type ClientEmailPayload = {
  [T in EmailType]: EmailTypePayload[T];
};

// Usage: Implementing 'sendEmail' function consuming the client email payload (as a parameter)
const sendEmail = <T extends EmailType>(payload: ClientEmailPayload[T]) => {
  // If we do something like this, we get the correct type safety/intellisense with respect to 'type'
  if (payload.type === EmailType.SendExtendedBookingLink) {
    payload.extendedBookingsDate;
  }
};

// Usage: Calling the function, passing a client email payload as argument (intellisense will work, checking for correct type interface with respect to 'type')
sendEmail({
  type: EmailType.SendExtendedBookingLink,
  customer: { name: "", surname: "" },
  extendedBookingsDate: "",
  bookingsMonth: "",
  organization: "",
});

// Usage: Constructing the payload for the function (typing the variable)
const emailPayload: ClientEmailPayload[EmailType] = {
  type: EmailType.SendBookingsLink,
  customer: { name: "", surname: "", secretKey: "" },
  organization: "",
};
```

- [ ] set the default templates:

- organization preferences should be initialised with the default values (default email template for each email type)
- we can create a cloud function (sort of a migration) to update existing organizations (in production) with the default templates
- the defaults can be stored as `data` somewhere (probably shared package so that we can use them for testing: client + functions packages)
- when updating the template (using `admin_preferences` route), we shouldn't allow for empty values

- [x] update the `sendEmail` function to consume the templates:

_some of the work is already there, but to reiterate and update:_
- we should update the `sendEmail` to check the email type (maybe add a validation for the type)
- we should check auth for the type (the caller should either be an admin, or have secret key, depending on email type)
- the payload should be validated (with respect to email type)
- if successful, the sendEmail should construct and write the email as `payload` to the process document, thus triggering the delivery

fadwamahmoud:
Fixes #681


ikusteu (changes requested):
I've commented on the PR, even though, some of the stuff I've commented should be changed altogether. 
As primary source of tasks, please use the updated #681 
After solving those, check the PR review comments, to fix the rest (and as future reference)

packages/functions/src/sendEmail/https.ts
```diff
@@ -19,27 +26,61 @@ export const sendEmail = functions
...
+            secretKey: emailPayload.customer.secretKey,
+          }))
+        )
       ) {
         throwUnauth();
       }
 
-      checkRequiredFields(email, ["to", "html", "subject"]);
+      // get email temp from org doc
+      const emailTemplate = await fetchOrganizationEmailTemplate(
```
ikusteu, line 46: Can we name this `getOrganizationTemplate` rather than `fetch` to conform to the naming we use so far.

packages/client/src/store/actions/bookingOperations.ts
```diff
@@ -222,27 +223,28 @@ interface sendICSFile {
 }
 
 export const sendICSFile: sendICSFile =
-  ({ icsFile, email, secretKey, displayName, name }) =>
+  ({ icsFile, email, secretKey, displayName = "displayName", name = "name" }) =>
```
ikusteu, line 226: `displayName` should most definitely not fall back to `"displayName"` (should probably be organization name)

ikusteu: Same goes for `name`

packages/client/src/__tests__/migrations.test.ts
```diff
@@ -305,4 +306,41 @@ describe("Migrations", () => {
...
+      "should populate default templates in existing organizations",
+      async () => {
+        const { organization } = await setUpOrganization();
+
+        const orgRef = adminDb.doc(`${Collection.Organizations}/org1`);
+
+        await orgRef.set({}),
+          await invokeFunction(CloudFunction.PopulateDefaultEmailTemplates)({
+            organization,
+          });
```
ikusteu, line 320: Can we just remove this test...out CI server is overworked anyway, the test is failing and is not really necessary

packages/client/src/__testSetup__/node.ts
```diff
@@ -32,7 +37,13 @@ export const setUpOrganization: SetUpOrganization = async (doLogin = true) => {
     // Create a new user in auth
     createUserWithEmailAndPassword(auth, email, pass),
     // Set given user as admin in org structure
-    orgRef.set({ admins: [email] }),
+    orgRef.set({
+      admins: [email],
+      emailTemplates: {
+        SendBookingsLinkTemplate,
+        SendCalendarFileTemplate,
```
ikusteu, line 44: I'm gonna go out on a limb and say this is the reason the tests below are failing. Just make it an object (like I've described) and it should be good to go.

packages/shared/src/types/firestore.ts
```diff
@@ -31,9 +31,14 @@ export interface OrganizationData {
    */
   emailNameFrom?: string;
   /**
-   * Template for reminder emails
+   * Templates for reminder emails
```
ikusteu, line 34: Not only for reminders

packages/shared/src/types/firestore.ts
```diff
@@ -31,9 +31,14 @@ export interface OrganizationData {
...
    */
-  emailTemplate?: string;
+  emailTemplates?: {
+    [name: string]: {
```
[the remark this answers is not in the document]
ikusteu: We can assume it will always be present

packages/shared/src/utils/organization.ts
```diff
@@ -0,0 +1,44 @@
+import { ClientEmailPayload, EmailType, HTTPSErrors } from "../index";
+export const interpolateEmailTemplate = (
+  template: string,
+  fieldValues: Record<string, string>
+): string => {
+  const placeholders = template.match(/\{\{ (.*?) \}\}/g) || [""];
```
ikusteu, line 6: I would like the whitespace to be optional (so `{{name }}` would also work)

ikusteu (changes requested):

packages/client/src/features/modal/components/ExtendBookingDateDialog/utils.ts
```diff
@@ -78,33 +79,48 @@ interface SendBookingsLink {
...
 
+      /** @TODO add bookingsMonth and extendedBookimgsDate to payload */
+      const emailPayload: Omit<
+        ClientEmailPayload[EmailType.SendExtendedBookingLink],
+        "organization"
+      > = {
+        customer: {
+          name,
+          surname,
+          email: email || "",
```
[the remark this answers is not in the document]
fadwamahmoud: it's optional all the way back in customer (which is what's passed to sendBookingsLink) so what do I make it fall back to?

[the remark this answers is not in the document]
ikusteu: Well...we can't send an email if one is not provided...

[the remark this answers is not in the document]
ikusteu: Besides, send bookings link button is disabled if there's no email present. Let's enforce better checks and assure that the email will, in fact be present in `sendEmail` function.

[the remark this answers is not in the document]
ikusteu: Ok...taking another look at it...let's just add `email!` non-null assertion and let the cloud function take care of it. It should throw a validation error if email is not present...

PR #698, opened 2022-12-30
Update/self registration

### LINKED ISSUE #696: Self registration debug/updates
@silviot and I have uncovered some bugs relating to the self registration functionality, as well as some required updates:

Sync the client side/server side validations for customer self registration form:
- [x] debug the difference
- [x] possibly show the backend validations in the UI (but syncing the form validations with db validations might just do it)

There was a bug with customer edit form: after the customer has self registered, using the (admin) UI to add a category was not allowed:
- [x] debug this

Misc:
- [x] when a new user is registered, the root route should redirect to the self registration form (like the /login for non registered users) instead of showing the not-authorised route
- [x] add some visual cue for non-confirmed athletes in admin view (make them greyscale/faded probably)
- [x] check for usage of `PhoneInput` (with default country) across the forms
- [x] enable self registration using phone number
- ~~[ ] create a data trigger updating user's `email`/`phone` in firebase auth~~
- [x] send an email to the admin each time a new customer is self-registered and should be verified

ikusteu:
Fixes #696 

Additionally:
- `medialCertificateExpiration` and `covidCertificateReleaseDate` are made optional
- name field from email signup is removed (as it's filled out at self registration step)


ikusteu:
@silviot I've already included the updates we've talked about (remove the data trigger - it's been rebased without it + aforementioned fields ☝️  are made optional) so this should be good to go as it is.

packages/client/src/components/auth/PrivateRoute.tsx
```diff
@@ -22,23 +23,29 @@ const PrivateRoute: React.FC<RouteProps> = (props) => {
   const isAuthEmpty = useSelector(getIsAuthEmpty);
   const isAdmin = useSelector(getIsAdmin);
   const isAuthLoaded = useSelector(getIsAuthLoaded);
+  const secretKey = useSelector(getBookingsSecretKey);
 
   switch (true) {
-    // display loading state until initial auth is loaded
+    // Display loading state until initial auth is loaded
```
silviot, line 29: Nitpick (and not related to the work in this PR): all cases in this `switch` block end with a `return` statement. The `switch` block could be removed and the `case` blocks could become `if`s. It would be more readable this way IMO. No need to change it, it's a small thing, but I wanted to share my thoughts about it.

packages/ui/src/Forms/CustomerProfile/CustomerProfileForm.tsx
```diff
@@ -70,6 +69,12 @@ const CustomerProfileForm = <P extends DefaultFormProps | SelfRegFormProps>({
     setIsEditing((isEditing) => !isEditing);
   };
 
+  // Disable email/phone if values already provided as first step of self-registration
+  const emailDisabled =
```
silviot, line 73: Nice touch!

silviot (approved):

PR #715, opened 2023-01-02
Split CustomerProfileForm into smaller blocks

ikusteu:
...to be used to construc different variants of the form


delasous (approved):
LGTM - v. clean separation

PR #723, opened 2023-01-06
Fix modal not closing on route chanage:

### LINKED ISSUE #712: Close the modal on route changes
Currently, when you navigate away from the page with modal open (by clicking the back arrow), the modal stays open, it would be better if redux could listen to router updates and close the modal each time the route changes.

ikusteu:
* Fixes #712 
* Add a listener (at App.tsx level) closing all modals on history route changes
* Add a cypress test testing said behaviour
* Additionally create 'AdminAria.NewCustomer' and assign to the '+' button in '/athletes' page

### DIFF
--- a/packages/client/src/App.tsx
+++ b/packages/client/src/App.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { Provider as ReduxProvider } from "react-redux";
-import { BrowserRouter } from "react-router-dom";
+import { BrowserRouter, useHistory } from "react-router-dom";
 import { getAuth } from "@firebase/auth";
 
 import CssBaseline from "@mui/material/CssBaseline";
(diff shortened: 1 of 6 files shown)


packages/client/src/App.tsx
```diff
@@ -17,28 +17,48 @@ import useConnectAuthToStore from "@/react-redux-firebase-auth/hooks/useConnectA
...
+    </ReduxProvider>
   );
 };
 
-export default App;
+/**
+ * An additional wrapper component, wrapping the app in a browser
+ * router to make the router context available at top level.
+ */
+const RouterWrapped = () => (
```
fadwamahmoud, line 58: Why wrap App separately, as opposed to wrap it in place (how it used to be)?

fadwamahmoud (approved):

packages/client/src/App.tsx
ikusteu: In order to be able to ‘useHistory’ in App the App component itself needs to have access to the router context and wrapping it in <Router> (like here), we create an instance of router context which is then accessible in the App “component”

PR #726, opened 2023-01-09
Link to customer profile from customer list items:

ikusteu:
* Clicking on a customer is BirthdayMenu (and BirthdayMenuDialog) links to customer profile
* Clicking on a customer in AttendanceCard links to customer profile
* Clicking on 'back' in customer profile links back to the previouse page (rather than /athletes page)
* The behaviour is tested using cypress tests
* Replaces some test-ids with aria labels (birthday menu, attendance card related)

### DIFF
--- a/packages/client/src/__testData__/testIds.ts
+++ b/packages/client/src/__testData__/testIds.ts
@@ -60,7 +60,3 @@ export const __create100Athletes__ = "create-athletes";
 // #region notificationButton
 export const __notificationButton__ = "notification-button";
 // #endregion notificationButton
-
-// #region birthdayBadge
-export const __birthdayMenu__ = "birthday-menu";
-// #endregion birthdayBadge
--- a/packages/client/src/__tests__/cloudFunctions.test.ts
+++ b/packages/client/src/__tests__/cloudFunctions.test.ts
@@ -176,7 +176,7 @@ describe("Cloud functions", () => {
       }
     );
 
-    test("should reject if no recipient provided", async () => {
+    testWithEmulator("should reject if no recipient provided", async () => {
       const { organization } = await setUpOrganization();
       const payload = {
         type: EmailType.SendBookingsLink,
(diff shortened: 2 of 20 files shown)


packages/e2e/integration/customer_profile_links.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+    );
+
+    // Click the 'back' button
+    cy.clickButton(i18n.t(ActionButton.Back));
+
+    // Check that we're back on the attendance page (by querying for slot interval, displayed on attendance card)
+    cy.contains(getSlotTimespan(slots["slot-1"].intervals));
+  });
+
+  it.only("links to the customer profile when clicking on a customer on birthday menu (and redirectsback to the given page on 'back' click)", () => {
```
fadwamahmoud, line 84: Why is this the only test to run?

packages/e2e/integration/customer_profile_links.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    // Click the 'back' button
+    cy.clickButton(i18n.t(ActionButton.Back));
+
+    // Check that we're back on the attendance page (by querying for slot interval, displayed on attendance card)
+    cy.contains(
+      i18n.t(OrganizationLabel.SettingsTitle, { organization }) as string
+    );
+
+    // Check again for saul in the birthday menu dialog
```
fadwamahmoud, line 123: I'm not sure I get the reason behind checking for the same functionality twice

fadwamahmoud (approved):

packages/e2e/integration/customer_profile_links.spec.ts
ikusteu: It's not the same functionality: one tests the dropdown, while other tests the dialog. The implementation may or may not be the same, but we're testing for behaviour rather than implementation.

packages/e2e/integration/customer_profile_links.spec.ts
ikusteu: Touche

PR #727, opened 2023-01-09
Enhancement/user readable notifications

### LINKED ISSUE #701: Create user readable notification messages
The notifications displayed are sometimes human readable (mostly success notifications, e.g. "Saul Goodman registered"), but for errors, we're, most of the time, simply displaying "Error" without getting into the specifics, to the point where even a developer would need to go to the network tab in the dev tools and infer the error.

Tasks:
- [ ] display human (non-tech person) readable error messages (for predictable errors), and keek the generic "Error" message to a minimum
- [ ] apply the same to success messages - most of which are already quite clear as successful actions are much more predictable than errors
- [ ] for each error message, also log the details to the console - this would help both in debugging, as well as, eventually in crash logs and so..

ikusteu:
- Fixes #701 
- adds error logging (in case an error notification is shown)
- replaces general error message with more precise error notifications (e.g. "Error updating customer: NAME SURNAME")

There are two preferences questions though:
- copy paste error is slightly generic "Error copying week/day", would you prefer to have it say error copying from WEEK to WEEK (od DAY to DAY)
- slot notifications say: slot updated, or slot update error (should we include more details)

_for both versions above ☝️ I didn't go into too much detail as I believe the notification would be too long_

### DIFF
--- a/packages/client/src/components/atoms/AttendanceCard/AttendanceCard.tsx
+++ b/packages/client/src/components/atoms/AttendanceCard/AttendanceCard.tsx
@@ -1,5 +1,6 @@
 import React, { useMemo } from "react";
 import { useDispatch } from "react-redux";
+import { Link } from "react-router-dom";
 
 import IconButton from "@mui/material/IconButton";
 import ListItem from "@mui/material/ListItem";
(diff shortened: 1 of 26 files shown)


fadwamahmoud (approved):

delasous (approved):

ikusteu:
Converted to draft. I have a version to pass all the checks, but would just like to clean up a bit, will do so tomorrow/ Sunday at the latest and then it can go in.

PR #728, opened 2023-01-10
Feature/shared jest setup

### LINKED ISSUE #708: Email templates follow-up
- [x] write tests for interpolation function - I've hand tested it in the node REPL, but it would be convenient to have tests for it, also, testing infrastructure should be set up for shared folder (it can just be copy/paste of test setup from other packages)
- [x] check email tests - now that the logic for email sending is opinionated (a few variants) we might want to extend the tests to test for different email type behaviour (those should probably be integration tests - utilising `jest-smtp`)

fadwamahmoud:
Fixes #708


delasous (approved):
LGTM

PR #730, opened 2023-01-16
Remove useSelector from utils

fadwamahmoud:


packages/client/src/features/modal/components/SendBookingsLinkDialog/utils.ts
```diff
@@ -78,12 +73,22 @@ interface SendBookingsLink {
...
+    name,
+    method,
+    email,
+    surname,
+    phone,
+    secretKey,
+    bookingsLink,
+    displayName,
+  }) =>
   async (dispatch) => {
```
ikusteu, line 92: This function returns a thunk, which has the following signature:
```typescript
interface FirestoreThunk {
  (
    dispatch: Dispatch // Redux dispatch wrapper
    getState: () => LocalState // Get the current state in the store
  ): Promise<void>
```
So, instead of passing a whole bunch of parameters, let's minimise them to the ones needed per function call, and have the function read from local store like so:
`const { displayName = getOrganization() } = getAboutOrganization(getState)[getOrganization()] || {}`

Also, let's avoid `__organziation__` as that's a default organization fallback and `getOrganization()` is more test-friendly way of using this (for cypress tests, where the organization is set per test basis). `getOrganization()` will still fall back to `__organization__` if no other organization is found

packages/client/src/features/modal/components/SendBookingsLinkDialog/__tests__/SendBookingsLinkDialog.test.tsx
```diff
@@ -27,8 +29,10 @@ jest
   .mockImplementation(mockSendBookingsLink as any);
 
 const mockDispatch = jest.fn();
+const mockSelector = jest.fn();
```
ikusteu, line 32: I don't like this and would rather test this `mockSelector` against an instance of the store (using `renderWithRedux`) and passing a store populated with some data, but this won't be necessary if you move the display name logic to the thunk itself.

ikusteu (changes requested):
I've left a couple of comments, but then I realised we don't need the `displayName` (and the complexity it brings at all...)

Can we, as a quick fix, simply replace the `displayName` in sms message with `${getOrganization()}`. This is not a huge shortcoming as the `displayName` will be interpolated when I'm done with SMS templates (in the same way it's interpolated in emails) and we wouldn't bring unnecessary complexity with passing of the `displayName` to the thunk, getting the display name using `useSelector` in the component, needing the store to tests out the `SendBookingsLinkDialog`, etc.

To be completely clear, my PR comments are valid as to how I'd normally approach this, but in this case, you can disregard them completely and simply use the `getOrganization()` (without the `displayName` altogether, rendering the comments obsolete 😄

ikusteu (approved):

PR #744, opened 2023-03-16
Extend email validations to include more special (but legal) characters

### LINKED ISSUE #737: Error when editing a self registered user
When users self register the admin needs to edit their profile to fill in categories. I was not able to do it in production using the UI. I had to use the firebase console.

ikusteu:
Fixes #737


silviot:
This problem is known to be harder than it looks at first sight.

I wonder if we should use a pre-baked regexp like the one you can find here: https://stackoverflow.com/questions/201323/how-can-i-validate-an-email-address-using-a-regular-expression

Otherwise a weird email address can cause similar problems in the future.

PR #758, opened 2023-05-04
Chore/repo maintenance 1

### LINKED ISSUE #762: Reinstate test reporting for unit tests
When migrating to Vitest (in #758), we've removed the `dorny/test-reporter` job in GH workflows as the reported output from Vitest differed from the Jest one and the action wasn't able to parse it. After quick try-and-fix updates, the issue was still there, so we've decided to move on with the core focus of the work (dependency updates) and reinstate the reporter and the reporting action.

ikusteu:
The main two drivers for updates applied in this PR were the wish to be able to use multiple package exports (as in node16 module resolution) and an error arising from `datebook` library (a dependency of ics calendar feature) where the resolution was picking up on ES code, rather than CJS.

The solution to both was to use node16 module resolution (allowing us to resolve to the required module standard, as per specification in `exports` field in packages `package.json`).

This, however led down the road of a proper dependency hell, where, due to outdated tooling, it was impossible to use the node16 resolution.

On the road to update the tooling/dependencies, I've:
- **Updated TypeScript to version 5** - this was a first step required in order to use `node16` (applied as `moduleResolution: nodenext` in tsconfig.json)
- Updated ESLint (and all ESLint related plugins) - this was to take `nodenext` resolution, set by TypeScript and not produce false errors
- **Updated Vite to version 4** - this was a requirement for migrating some of the dependencies to newer version using Vite (instead of Webpack/Babel), which, in turn works out of the box with TypeScript and automatically applies the config specified in tsconfig.json, without, like Webpack, requiring additional (and painful) configuration
- **Updated Storybook to the latest version** - former Storybook versions used Webpack, which, in my opinion, is an outdated tool and provides no leg up compared to Vite. Luckily, latest Storybook version uses Vite out of the box, but it required the latest version of Vite in order to work properly
- **Replaced Jest with Vitest** - for the same reason as Storybook - Jest is using Babel to parse TS files, whereas Vitest (as the name implies) uses Vite (integrating seamlessly with TypeScript), one caveat though - Vitest breaks on older versions of V8 (shipped with Node16)
- **Updated Node version to 18** - It is LTS and ships with updated V8 version, fixing problems Vitest was having
- **Updated Firebase (and all Firebase modules)** - when using `nodenext` resolution, the older versions (the ones we were using) of Firebase cause TS errors as they didn't have a specified `exports` field (and probably build process dodn't support it), signalling to TS resolver which files to pick up, as well as where the type declarations should be (for `nodenext` resolution)
- **Created `firestore/compat` utils, leveraging `variants` library (implementing discriminated unions for TypeScript)** - this was used to provide single api for `doc`, `getDoc`, `collection`, etc. working with variants, instead of implementation specific objects to provide single api compatible with client `@firebase/firestore`, server `@google-cloud/firestore` and "compat" `@firebase/rules-unit-testing` (this came especially handy in tests, where we use combination of the aforementioned firestore SDK implementations for setup/testing/client code being tested)
- **Updated `firebase-tools` (emulators) to the latest version** - our version 9.x wasn't really working with Node18 and (while using Node16) had some behaviour bugs integrating with our (latest) Firebase SDKs
- **Updated test reporting** - Vitest is slightly different than jest in terms of test report (it constructs a Java-like Junit report, whereas for jest, we've used jest-junit) Fixes #762 
- **Added support for Vitest UI** - a nice and powerful tool providing us with in-browser UI dashboard while running tests

As the tooling updates finally came together, I've applied the updates to the way our code is organised to reduce duplication (as well as test out that the given structure will work - that structure was mentioned as on of the two drivers for this update):
- moved all test ids (and other test data) to the `testing` package - able to be shared across our packages (thus removing some existing duplication)
- created a separate build of `shared` package, imported using `@eisbuk/shared/ui`, providing shared ui-related artefacts (like shared hooks, UI only enums, etc.)

Lastly, while refactoring the tests to make them work with Vitest, I've encountered some messy tests and implementations, so I've removed flakiness and updated both with slightly cleaner code (and much needed comments as inline documentation)

The work is finally stable, but is not perfect: I've opened #763 #761 to address flakiness in E2E tests and unit tests


silviot (approved):
Thanks! I think you did an amazing job here.

For me this has been one more occasion to remember to always try not to add a dependency, if possible.

I didn't really go into the code, but I did try the app and all the functionality I touched (hopefully most of what is there) worked perfect.

PR #760, opened 2023-05-24
Bump all mentions of Node.js versions to 18.x.x

ikusteu:
This should, hopefully, work without fault (as we've merely updated to newer, current LTS, version of Node.js) and should enable other updates: e.g. Vitest uses ArrayBuffer feature, which fails on older V8, which is supposedly fixed in the newer versions of Node


ikusteu:
Included in #758

PR #766, opened 2023-05-30
Vitest matrix/initial

### LINKED ISSUE #763: Create a unit test matrix
Just as we have a cypress-matrix, there to pinpoint flakiness in cypress tests, I feel we should also create on such workflow for unit tests.
The issue I've experienced, while working on #758 was that unit tests, even though passing locally, were sometimes flaky in CI. The reasons for failure were quite weird and hinted to firebase emulators not being able to handle the load from all of the test cases interacting with the emulators in parallel. I did make progress on that front and have made some sensible updates, where possible, but a weird issue still arises from time to time. 
For that use case, we should create the aforementioned workflow and use it to pinpoint/debug weird behaviour in our unit tests.

ikusteu:
Fixes #763


ikusteu:
Merging this in as it will prove useful for flaky tests, features no braking changes and the only failing checks are from flakiness while running the matrix

PR #777, opened 2023-07-06
Prominent sign in methods

### LINKED ISSUE #772: Make phone and email link login methods more prominent: 2 pts
Athletes should "see" only these two methods.
The other methods should still be available, but only for people who know they're there.
People seeing this for the first time should only decide between phone and email, so we should present these two prominently (it's ok as they are now) and the other two "on the side of the stage". I was thinking as small circles in the bottom right, or even on a different route like `login-advanced`. They're useful for testing/power users who know they're there.

ikusteu:
Fixes #772 

Emphasise 'sign in with phone' and 'sign with email link' as main sign in methods:

* Make the 'sign in with google' and 'sign in with email and password' buttons smaller (icon only) on login screen
* Add Key and Google icons to 'svg' package
* Update english translation for email signin to 'sign in with email and password'
* Update related e2e tests


silviot (approved):
LGTM

PR #782, opened 2023-07-08
Remove all mentions of deprecated categories

### LINKED ISSUE #720: Eradicate deprecated categories: 2 pts
We've switched categories a while back (#478). At the time we've provided backwards compatibility (and a migration), but I think by now we should switch to the "new" categories fully and remove unnecessary confusion and unnecessary code.

- [ ] Check if the migration has been ran in the production deployments (and if not run it)
- [ ] Remove the `DeprecatedCategory` from the code

ikusteu:
Fixes #720


silviot (approved):
LGTM

PR #785, opened 2023-07-09
Feature/shortened names

### LINKED ISSUE #360: Create logic for displaying customers with multiple names/surnames
Most of our UI proportions are tested with long names/surnames, but this didn't include customers with multiple names/surnames. We should create a logic to shorten the names in such case (visually, without affecting the customer's data in the db).

My suggestion is to display up to two names, first full and second just the first letter.
Example:
```typescript
const name = "Sterling Malory"
const surname = "Archer"
const displayName = getDisplayName(name, surname) // "Sterling M. Archer"
```
Similar logic would be applied to last name only the latest would take presedance:
```typescript
const name = "Pablo Emilio"
const surname = "Escobar Gaviria"
const displayName = getDisplayName(name, surname) // "Pablo E. E. Gaviria" 
```
The rationalization for the last name logic comes from the edge case of having prefixed surnames, such as "van Beethoven", "von der Leyen", etc.. In that case, it would make sense to omit all lowercase words ("von", "der"), resulting in "clean" names such as "Beethoven" and "Leyen", while in case where the prefixes are uppercased (intentionally, or by mistake), the surname would look like this: `V. D. Leyen`

ikusteu:
Fixes #360


silviot (approved):
Can you do one last change?

A name like `Vittorio De Sica` is currently rendered as `Vittorio D. Sica`. That's a very common instance, and there's no point in turning the `e` into a period. `Robert De Niro` is another istance of this.

PR #786, opened 2023-07-10
Athletes approval

### LINKED ISSUE #775: Improve usability of athlete "approval": 5 pts
When athletes self subscribe they need to be approved by an admin.

Unapproved athletes are shown in grey on the athlete page. It should be easier for an admin to answer the questions:
* how many athletes are there to approve?
* Are there any?
* How do I approve them?

I think a good way to improve on the current status could be:

* Display an icon at the top when there are athletes to approve
* A click on the icon should bring the user to a view (possibly at route `/athletes-to-approve`) where only athletes needing approval are shown.

fadwamahmoud:
Fixes #775


silviot (changes requested):
I tried to remove a user instead of approving them. They're still listed amongst the to-be-approved.
Please change the list so that it omits deleted users.

fadwamahmoud:
> e the list so that it omits deleted users.

I've tried reproducing it but the user is being removed and I've written one more cypress test for that scenario and it passes. Would you mind sharing a screencast? @silviot

silviot (approved):

PR #791, opened 2023-07-12
Add loading spinner to submit button on self reg

### LINKED ISSUE #787: Improve `/self_register` usability - sumbission indication: 1pts
When the user clicks "Save" on the self register form they need to wait for form submission. This will typically take a long-ish time, given the nature of the deployment (firebase functions "go to sleep" when not used).

As a user who just clicked "Save" I see no indication that I actually clicked. It's frustrating.

fadwamahmoud:
Fixes #787

### DIFF
--- a/packages/client/src/pages/self_register/index.tsx
+++ b/packages/client/src/pages/self_register/index.tsx
@@ -54,7 +54,7 @@ const SelfRegisterPage: React.FC = () => {
   // to more correctly control the 'isSubmitting' state
   const submitForm: Parameters<
     typeof CustomerForm.SelfReg
-  >[0]["onSave"] = async (values, { setErrors }) => {
+  >[0]["onSave"] = async (values, { setErrors, setSubmitting }) => {
     const { secretKey, codeOk } = await customerSelfRegister(values)(
       dispatch,
       getState,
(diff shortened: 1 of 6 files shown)


packages/e2e/integration/athlete_self_register.ts
```diff
@@ -91,4 +91,39 @@ describe("Athlete self registration", () => {
       t(Alerts.ContactEmail, { email: __emailFrom__ }).replace(/<\/?a.*>/g, "")
     );
   });
+
+  it("checks for a spinner on the submit button while the submit request is processing", () => {
```
ikusteu, line 95: it "shows a spinner while the submit request is being processed". `it` being the code you're testing, not the test itself.

ikusteu: > "It puts the lotion in the basket"
😂

packages/client/src/pages/self_register/index.tsx
```diff
@@ -68,6 +68,7 @@ const SelfRegisterPage: React.FC = () => {
       setErrors({
         registrationCode: t(ValidationMessage.InvalidRegistrationCode),
       });
+      setSubmitting(false);
     }
```
ikusteu, line 72: Ok, so I would wrap this into a try { ...all of the code } finally { setSubmitting(false) }. That way, whatever happens, the submission will be done (as far as the submitting state is concerned)

ikusteu (approved):
A click through test works. Left some minor suggestions.

packages/e2e/integration/athlete_self_register.ts
silviot: image

silviot (approved):

PR #794, opened 2023-07-16
Create deleted page and redirect

### LINKED ISSUE #789: Add a "Deleted" page for deleted athletes: 2pts
When we delete an athlete we don't currently remove their record, but rather mark it as deleted.
Our code takes this into account, except for the athlete themselves: they can still login and won't notice any problem.

Implement a view that deleted athletes will see on every route. It should have a text that reads "Your account has been removed. If you believe this was a mistake please email <admin@email>".

fadwamahmoud:
Fixes #789


ikusteu (changes requested):
.

ikusteu:
Merging this in as the tests are passing locally and I'd really like to move this along. If the problem persists, will open a ticket and debug

PR #803, opened 2023-09-07
Fix/error removing athletes

fadwamahmoud:


ikusteu (changes requested):
Let's first merge the #794 as this already includes that change

Issue #804, opened 2023-09-08
Athletes can't book. They get "Firestore permission denied"


silviot:
To make this work in production I replaced the current `firestore.rules` file with these contents:

```
rules_version = '2';
service cloud.firestore {
  match /databases/{database}/documents {
    match /{document=**} {
      allow read, write: if true;
    }
  }
}
```
I did it from the console. I'll update code in the repo too.

fadwamahmoud:
There are two culprits behind the bug:

    A few of the slots that can't be booked have data discrepancies (they have extra props which shouldn't be there className"w-full" and selected:false)
    The majority of this bug is that the slots from September and October are mixed up (they're both showing in both months)
    So when an athlete tries to book a slot while they're in September and that slot actually belongs in October it'll show that error.

fadwamahmoud:
Root cause is slotsByDay having multiple dates with the same slots
![Untitled(1)](https://github.com/eisbuk/EisBuk/assets/41344489/1a0bfb78-ee0f-47e6-a0f2-c82fa186e4bd)
![Untitled](https://github.com/eisbuk/EisBuk/assets/41344489/7676a92e-ab7c-43c0-849a-bad3082dcfc5)

fadwamahmoud:
I have a hunch that this was caused by the copy/paste bug somehow since the duplicate slots are exactly the same, but I can't go any further than that. I suggest we fix the data manually since it's just these specific dates that cause the error (new slots can be booked fine) by deleting one of each duplicate set of slots and creating the deleted slots again. @silviot @ikusteu

silviot:
@fadwamahmoud So the attempted (and failed) copy/paste operations did result in changes of the database, if I understand correctly.

If that is the case I think we can manually remove the faulty records from production, but only as long as we save the existing bad records to track down the original problem.

A side note: a relational database would have prevented that. Typically a web server would rollback its current transaction and any change performed in the current request would be discarded.

fadwamahmoud:
The copy/paste guess is just a hunch because I couldn't find another reasonable explanation after failing to find a bug; I could be totally wrong @silviot 
@ikusteu feel free to take a look maybe this needs a fresh set of eyes

ikusteu:
Partially related to this: I've opened up [this discussion](https://github.com/eisbuk/EisBuk/discussions/809) related to datamodel and complex data flows.

ikusteu:
[...]

PR #805, opened 2023-09-08
Disable firestore rules as a workaround fix for 804

silviot:
I understand this will break tests.

But until this is fixed the production firestore rules should be so that they allow booking, and for now the easy solution is to leave the all-permissive ones in place.

To clarify, if the fix for #804 is the next thing to be merged, then this PR can be closed and not merged.
If instead something else needs to go in before the fix for #804 then either the person doing the merge needs to manually go to the production firebase console and nuke the rules there, or this PR should be merged.


fadwamahmoud:
I'm taking a look at #804 @silviot @ikusteu

fadwamahmoud:
There are two culprits behind the bug:
1. A few of the slots that can't be booked have data discrepancies (they have extra props which shouldn't be there `className"w-full"` and `selected:false`)
2. The majority of this bug is that the slots from September and October are mixed up (they're both showing in both months)
So when an athlete tries to book a slot while they're in September and that slot actually belongs in October it'll show that error. I'm not sure why that happened but will look into it + how to fix it tomorrow @silviot

ikusteu (approved):
I say we merge this and then reinstate the rules when (or after) the #804 is worked on

PR #810, opened 2023-09-10
Some copy/paste related fixes:

### LINKED ISSUE #797: Copy paste problem
Trying to copy/paste in production yields an error. The console shows `FirebaseError: Function doc() cannot be called with an empty path.`.

ikusteu:
* In firestore compat `doc` function, client variant: if doc path is not provided (we wish a new server-assigned id), pass only the parent node (to prevent empty path crashes)
* In firestore compat `batch` function, bind `this` to returned `commit` methods (for both client and server variant)

Fixes #797

### DIFF
--- a/packages/client/src/utils/firestore/compat.ts
+++ b/packages/client/src/utils/firestore/compat.ts
@@ -94,12 +94,14 @@ export const doc = (
   return match(node, {
     [FirestoreEnv.Client]: ({ instance }) =>
       FirestoreDocVariant.client({
-        // Even though 'doc' (aliased as 'clientDoc' here) works on both firestore and collection instanceerences,
+        // Even though 'doc' (aliased as 'clientDoc' here) works on both firestore and collection references,
         // it doesn't work well on unions because of slightly different type overloads, hence the typecast.
-        instance: clientDoc(
-          instance as ClientFirestore,
-          pathSegments.join("/")
-        ),
+        instance: docPath.length
+          ? clientDoc(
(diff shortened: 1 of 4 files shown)


packages/client/src/utils/firestore/compat.ts
```diff
@@ -94,12 +94,14 @@ export const doc = (
...
+        // Even though 'doc' (aliased as 'clientDoc' here) works on both firestore and collection references,
         // it doesn't work well on unions because of slightly different type overloads, hence the typecast.
-        instance: clientDoc(
-          instance as ClientFirestore,
-          pathSegments.join("/")
-        ),
+        instance: docPath.length
+          ? clientDoc(
+              instance as ClientCollectionReference,
+              docPath.length ? docPath : undefined
```
silviot, line 102: Isn't this condition redundant? We're in a conditional branch where `docPath.length` is `true`. I'd replace `docPath.length ? docPath : undefined` with `docPath`.

silviot (approved):
I don't fully understand the fix, but I tried it and it works. I left a comment about a minor adjustment that IMO needs to be made.

packages/client/src/utils/firestore/compat.ts
ikusteu: Yup, good catch. This is a leftover from previous attempt (using only `clientDoc(instance, docPath.length ? docPath : undefined)`, without the first check) which didn't seem to work

PR #812, opened 2023-09-10
Account for interval time when sorting customer calendar view 

### LINKED ISSUE #799: Sort entries in athlete personal calendar view
When athletes go to their `Calendar` tab they see the lessons they booked.
They're sorted OK by day. But sometimes two slots in the same day (it's not uncommon for ice and off-ice sessions to be one after the other) are not sorted by time.

ikusteu:
Fixes #799


packages/client/src/store/selectors/bookings/__tests__/bookings.test.ts
```diff
@@ -349,5 +353,108 @@ describe("Selectors ->", () => {
         },
       ]);
     });
+
+    test.only("should sort the slots (by date, and by time intraday)", () => {
```
silviot, line 357: I guess this `only` was not supposed to end up in the final commit.

packages/client/src/store/selectors/bookings/slots.ts
```diff
@@ -133,14 +133,13 @@ export const getMonthEmptyForBooking = (state: LocalStore): boolean => {
   return isEmpty(getSlotsForCustomer(state));
 };
 
-type BookingsList = Array<SlotInterface & { interval: SlotInterval }>;
-type BookedAndAttendedList = Array<
-  SlotInterface & { interval: SlotInterval } & { booked: boolean }
->;
+type BookingsEntry = SlotInterface & {
+  interval: SlotInterval;
+  booked: true;
```
silviot, line 138: Shouldn't this be `boolean`?

packages/client/src/store/selectors/bookings/__tests__/bookings.test.ts
ikusteu: Not at all...on it

PR #826, opened 2023-09-16
Make the slot form action buttons always visible 

### LINKED ISSUE #802: Make the `Save`/`Cancel` buttons on the slot edit form always visible.
Currently the buttons at the bottom on the form appear below the fold most of the time.
They should be always visible at the bottom of the modal.

See how, in the screencast below, I got confused and clicked "Add interval" at around 40 seconds. My eyes were looking for the "save" button and were assuming it was visible without scrolling. I think it's a reasonable expectation to have.

[Screencast from 06-09-23 11:36:44.webm](https://github.com/eisbuk/EisBuk/assets/396425/8f57008a-33ee-44c8-9713-77c55f6e504e)

ikusteu:
Fixes #802


silviot (approved):
LGTM

PR #831, opened 2023-09-18
Feature/error boundary

### LINKED ISSUE #808: Add error boundaries to views
Problems like #807 are exacerbated by the fact that the whole app crashes. Instead, when a similar error happens, only the part of the view with an error should be unavailable. The rest of the app should stay usable.

fadwamahmoud:
Fixes #808


packages/shared/src/ui/enums/routes.ts
```diff
@@ -4,6 +4,7 @@ export enum Routes {
   Unauthorized = "/unautorized",
   SelfRegister = "/self_register",
   CustomerArea = "/customer_area",
+  ErrorBoundary = "/error_boundary",
```
silviot, line 7: I'd love this URL to be super-obvious. The current one is very obvious if you look at it in the context of this PR, but https://igoriceteam.web.app/error_boundary wouldn't suggest to me what it does.

I propose `/do_trigger_an_error` for this view path.

silviot (approved):
I left a request for a small change in the route name.

Apart from that, here's my ignorant "Seal of approval": I clicked through this and skimmed the code. LGTM

image

packages/client/src/pages/attendance/index.tsx
```diff
@@ -74,10 +82,10 @@ const AttendancePage: React.FC = () => {
     { collection: OrgSubCollection.SlotsByDay },
   ]);
 
+  const calendarNavProps = useDate();
   const [view, setView] = useState<keyof typeof viewsLookup>(Views.ByDay);
-  const AttendanceView = viewsLookup[view];
+  const AttendanceView = viewsLookup[view]([calendarNavProps]);
```
ikusteu, line 87: I don't really understand why the `resetKeys` is necessary (my guess is it has something to do with the memoization).

However...this is React, not svelte. By that I mean...every time a rerender happens, this function will be called, to retrieve the view function. This, in turn, means that the returned JSX will be fully rerendered (as the component has changed). This is something you pick up from years of using React, I would like to not think about it and I wouldn't want anybody else to have to think about it, so let me summarise:

It would be better to simply pass `reserKeys` to the view component, instead of having the component be a higher order function, returning a component, like so:

```jsx
[Views.ByDay]: ({resetKeys}: {resetKeys: Array<any>}) =>
    (
      <LayoutContent>
        <ErrorBoundary resetKeys={[resetKeys]}>
          <ByDayView />
        </ErrorBoundary>
      </LayoutContent>
    ),
```

And then you simply do:

```jsx
<Layout>
  <Calendar {...calendarProps} />
  <AttendanceView resetKeys={resetKeys} />
</Layout>
```

ikusteu (changes requested):

packages/client/src/pages/attendance/index.tsx
ikusteu: Going back to this...what does the `resetKeys` do? I removed it completely and didn't observe any different behaviour...

fadwamahmoud: ResetKeys trigger a reset of the error boundary when they change. When you removed it did the error boundary disappear and non-error view get rendered when you switched between dates?

ikusteu:
I'm not sure this works...I've tried adding this to `packages/client/src/pages/slots/index.tsx:138`:

```diff
<ErrorBoundary resetKeys={daysToShow}>
+ {Object.keys(null).map(() => "unreachable")}
  {daysToShow.map((dateISO) => {
```
And I got this error:
```
...full stack trace

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries.
```

silviot:
There's a couple of problematic dates in production. I'll merge this since it's immediately useful.

PR #835, opened 2023-09-20
create custom ErrorBoundary to support render prop

fadwamahmoud:


ikusteu:
This seems to be a TEMP branch and all relevant updates included in #836 
If that's not the case, please reopen the PR

PR #838, opened 2023-09-24
Feature/sms templates

### LINKED ISSUE #700: Use emailTemplates (or similar templates) for SMS
We're also using SMS to send notifications/data to the customers, so `sendSMS` could be rewritten to behave in the same way as `sendEmail`:
- data required to construct/interpolate a certain SMS type is passed to an endpoint
- endpoint performs validations and interpolation
- endpoint stores (completed) SMS payload to the sms delivery queue
- SMS deliver function is used only for delivery, not for validation/constructing

We should check the current email types and probably reuse both the types and templates to construct an sms/email

ikusteu:
Fixes #700


silviot (approved):
I was unable to go through the code. I tried the functionality, and I really like how it looks now. I think it would make sense to not include `OrganizationName` in the template variables (admins will find hard coding the string easier IMO) but there's no need to make changes to this PR. Merging!

PR #847, opened 2023-10-10
Feature/separate slot types attendance month

### LINKED ISSUE #842: Separate ice and non-ice in the montly attendance view
Each athlete should have two lines: one for ice and one for non-ice. The two categories should be accounted separately.

ikusteu:
Fixes #842 

It's a bit rough around the edges - customer name is displayed twice:
- once for ice attendance rows (booked and delta)
- once for off-ice attendance rows (booked and delta)

As far as the data flows and correctness go: It seems fine to me (click-through) and I've tested it to extent that seemed reasonable


packages/client/src/store/selectors/attendance/attendanceVariance.ts
```diff
@@ -1,63 +1,88 @@
...
+type Slots = Record<string, SlotInterface>;
 
-  return attendanceTableData;
+export const processAttendances = (
+  attendance: Attendance,
+  slots: Slots,
+  customers: Customers,
+  month: string
+) => {
+  const final = wrapIter(Object.entries(attendance))
```
silviot, line 41: Is it too much to turn this into an arrow function and remove 
```
{
  const final =
```
and remove the final `return final;}`?

silviot (approved):
Does not also fix #843 unfortunately, but looks good (enough) to me.

delasous (approved):

silviot (approved):

PR #850, opened 2023-10-17
Feature/list athletes on booked intervals

### LINKED ISSUE #837: Prevent editing/removing intervals that have been booked
As mentioned in #834 the app breaks when an interval is changed or removed if it has been booked.
Simplest solution is to implement the same logic used for deleting slots.

fadwamahmoud:
Fixes #837


silviot (approved):
I didn't look into the code. LGTM

PR #851, opened 2023-10-18
Changes to the printable attendance screen

### LINKED ISSUE #844: Change printable attendance
The `/attendance_printable` view needs these changes:

* Remove header
* ~Remove signature column~
* ~Add empty row after each athlete row to allow for signature~
* Sort athletes by last name (in each booked slot/period)
* Align athletes last names initials
* Make last names bold

fadwamahmoud:
Fixes #844


silviot (changes requested):
Nice! A couple more changes:
* remove both the "note" columns - the notes athles can write on a booked slot have a "personal" label, so we should not include them here
* the header still takes a bit too much space. Thinking more about this, I think it should be `Organization name - YYYY-MM-DD` to include the date.
image

silviot:
@fadwamahmoud I was giving this a final look when it hit me: as it is there will never be page 2. If there are too many athletes the printed page will include a scrollbar. But users don't usually know how to use scrollbars on papers, so they'll miss the last athletes in the list.

image

We need to make sure all names are printed (on more than one page if necessary) before we can merge this.

silviot:
One more additional change: please swap first and last name. Write athlete names as "Lastname Firstname".

silviot:
Also add back in (I'm sorry for this back and forth - I'll make sure it doesn't happen again):

* trainer column
* notes column (one is enough)
Both columns are meant to be filled in with pen on paper, and printed blank.

silviot (approved):

PR #855, opened 2023-10-19
Turn the flapping test into a soft failure

silviot:
The `should create attendance entry for booking and not overwrite existing data in slot` test often fails, and wastes our time. This PR turns it into a non-failure.
The idea is that people working on that part of the code should revert this (temporarily so would be ok) to get test coverage back.


silviot:
Superseded by c2a7c050549ab8216d19c63be3bd3160f713c30d

PR #857, opened 2023-10-21
Feature/db sanity check 1

ikusteu:
Started with respect to recent issues with production db mismatches, this is the initial implementation of db sanity checks:
- checks for mismatches between `slots` and `attendances` collections
- additionally - provides a more convenient way of running the app locally with production data

_note: more tickets will stem from this work, tackling db checks issue, as well as backup-restore enhancements_


packages/functions/src/checks/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  const slots = await orgRef
+    .collection(OrgSubCollection.Slots)
+    .get()
+    .then((snap) => snap.docs);
+  const attendances = await orgRef
+    .collection(OrgSubCollection.Attendance)
+    .get()
+    .then((snap) => snap.docs);
+
```
silviot, line 28: Shouldn't we also go through `OrgSubCollection.Bookings`? I'm 99% sure this work is already planned, but I mention it anyway, in case we're in that 1% case.

silviot (approved):
I left a couple of comments that need discussion but the general approach looks good. I tried the function locally and worked gread (despite a couple of emulators hiccups).

silviot (approved):

PR #885, opened 2023-11-13
Feature/booking stats data trigger

### LINKED ISSUE #845: Athlete list needs to include booking stats
The `/athletes` view needs to include information about booked hour totals for the current and next month.
One way to do this could be having each athlete on a row and including totals next to the name/avatar.

fadwamahmoud:
Fixes #845


packages/functions/src/dataTriggers.ts
```diff
@@ -461,3 +466,116 @@ export const createAttendedSlotOnAttendance = functions
...
+      .doc(organization)
+      .collection(OrgSubCollection.Bookings)
+      .doc(secretKey);
+
+    // Fetch the booking document
+    const bookingDoc = await bookingRef.get();
+    const { id: customerId } = bookingDoc.data() as CustomerBookings;
+
+    // Fetch documents from a subcollection of the booking
+    const bookedSlotsRef = bookingRef.collection("bookedSlots");
```
silviot, line 494: Can we skip the `bookedSlotsRef` name and do `const bookedSlotsSnapshot = await bookingRef.collection("bookedSlots").get();` on single line?

packages/functions/src/dataTriggers.ts
```diff
@@ -461,3 +466,116 @@ export const createAttendedSlotOnAttendance = functions
...
+    const bookedSlotsSnapshot = await bookedSlotsRef.get();
+
+    const bookedSlots: DocumentData = {};
+    bookedSlotsSnapshot.forEach((doc) => {
+      Object.assign(bookedSlots, { [doc.id]: doc.data() });
```
silviot, line 499: Is there a reason not to write this as `bookedSlots[doc.id] = doc.data();`?

packages/functions/src/dataTriggers.ts
```diff
@@ -461,3 +466,116 @@ export const createAttendedSlotOnAttendance = functions
...
+    const nextMonthSlots = (
+      await db
+        .collection(Collection.Organizations)
+        .doc(organization)
+        .collection(OrgSubCollection.SlotsByDay)
+        .doc(nextMonthStr)
+        .get()
+    ).data() as SlotsByDay;
+
+    const stats = Object.entries(bookedSlots).reduce(
```
silviot, line 529: I see a great opportunity to reduce the size of this function (that starts online 475 and is passed to `onwrite`) by extracting the `stats` calculation outside.

To expand: this function currently takes care of two things. It fetches the data it needs and then combines it to the shape it needs. I'm suggesting to extract the second task into a named function, with its own (small) documentation. The new function should receive `currentMonthSlots` and `nextMonthSlots` and return a `stats` object. This new function will be testable without emulators.

packages/functions/src/dataTriggers.ts
fadwamahmoud: No, yours is more readable, actually. They do the same thing but mine was just out of habit.

packages/functions/src/utils.ts
```diff
@@ -332,3 +340,105 @@ export const validateJSON = <T extends Record<string, any>>(
...
+ * @param {string | null} interval - String slot interval
+ * Converts a string slot interval to a number e.g:
+ * `null => 0`;
+ * `"21:00 - 21:20" => 0.5`;
+ * `"16:00 - 17:00" => 1.0`;
+ * `"22:00 - 23:30" => 1.5`;
+ * `"22:00 - 24:00" => 2`;
+ * `"22:20 - 24:00" => 2`;
+ */
+export const calculateIntervalDuration = (interval: string | null) => {
```
silviot, line 356: We already have logic to calculate an interval duration. Duplicating this piece of logic enables some bugs (misalignment of what's seen on one screen and what's seen on another one). We can easily avoid them by keeping a single implementation of this funciton throughout the codebase.

I suggest to factor out [the existing function](https://github.com/eisbuk/EisBuk/blob/master/packages/ui/src/AttendanceSheet/utils.ts#L11C14-L32) into the [shared package](https://github.com/eisbuk/EisBuk/blob/master/packages/shared/src/utils/slots.ts) and import it from there into both `functions` and `ui`. A little change will be needed, since the current function returns a string for display, but here we need a value to calculate a grand total, but the core logic needs to be kept in a single place, and not duplicated.

packages/client/src/__tests__/dataTriggers.test.ts
```diff
@@ -674,4 +675,177 @@ describe("Cloud functions -> Data triggers ->", () => {
...
+            .doc(getSlotDocPath(organization, `${dateThisMonth}-10`))
+            .set(slotThisMonthOffIce),
+          adminDb
+            .doc(getSlotDocPath(organization, `${dateNextMonth}-10`))
+            .set(slotNextMonthOffIce),
+          adminDb.doc(getCustomerDocPath(organization, saul.id)).set(saul),
+        ]);
+
+        await Promise.all([
+          await waitFor(async () => {
```
ikusteu, line 723: This `await waitFor` is unnecessary as `waitFor`returns a promise, one which we can `await`, however, `Promise.all` already awaits all promises in the array.

packages/client/src/__tests__/dataTriggers.test.ts
```diff
@@ -674,4 +675,177 @@ describe("Cloud functions -> Data triggers ->", () => {
...
+            const snap = await adminDb
+              .doc(getCustomerDocPath(organization, saul.id))
+              .get();
+            expect(snap.data()).toEqual(saul);
+          }),
+        ]);
```
ikusteu, line 729: I understand the additional check (just-to-make-sure), but when we store saul into customers collection (in the previous `Promise.all` block), the saul will be there (we need to do this check only when we depend on a data trigger being ran in the background and want to verify it had completed)

packages/client/src/__tests__/dataTriggers.test.ts
```diff
@@ -674,4 +675,177 @@ describe("Cloud functions -> Data triggers ->", () => {
...
+              getBookedSlotDocPath(
+                organization,
+                saul.secretKey,
+                `${bookedSlotThisMonthIce.date}-9`
+              )
+            )
+            .get();
+
+          expect(snap.data()).toEqual(bookedSlotThisMonthIce);
+        });
```
ikusteu, line 802: Same as when adding customer: If we explicitly store the doc, the doc is there when the `.set` promise resolves, no need for additional check.

packages/shared/src/utils/slots.ts
```diff
@@ -28,3 +28,45 @@ export const getSlotTimespan = (
   // return time string
   return `${startTime} - ${endTime}`;
 };
+
+// #region CalculateInterval
+export const calculateIntervalDurationInMinutes = (interval: string | null) => {
+  if (interval === null) {
+    return 0;
+  }
```
ikusteu, line 36: Finally 👌 
I've been meaning to move this here forever now, but always seem to forget, tnx

packages/ui/src/AttendanceSheet/utils.ts
```diff
@@ -8,24 +8,19 @@ enum BookingDuration {
...
 
-export const calculateIntervalDuration = (
-  startTime: string,
-  endTime: string
-): BookingDuration => {
-  const luxonStart = DateTime.fromISO(startTime);
-  const luxonEnd = DateTime.fromISO(endTime);
-  const { minutes } = luxonEnd.diff(luxonStart, ["minutes"]);
+export const getIntervalString = (interval: string): BookingDuration => {
+  const durationInMinutes = calculateIntervalDurationInMinutes(interval);
```
ikusteu, line 12: Let's not use `getIntervalString` here. That would imply you'd get something like: `"18:00-19:00"` string.
Maybe call this `getDurationLabel` or `getIntervalLabel` if you want to change it from `calculateIntervalDuration` (and I see as point in doing so)

ikusteu (changes requested):
- the migration function was not updated
- I've pointed out some wrinkles here and there

As for the migration / data trigger discrepancy, I would recommend extracting the shared functionality in a single function and then calling the same function (with different parameters) from both the trigger and the migration

packages/functions/src/migrations.ts
```diff
@@ -213,6 +217,90 @@ export const clearDeletedCustomersRegistrationAndCategories = functions
...
+        const thisMonthStats = getCustomerStats(
+          bookedSlots,
+          currentMonthSlots,
+          currentMonthStr
+        );
+        const nextMonthStats = getCustomerStats(
+          bookedSlots,
+          nextMonthSlots,
+          nextMonthStr
+        );
```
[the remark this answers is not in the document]
fadwamahmoud: No, it's the new structure I just assumed that when the migration is used it should calculate stats for current and following months

fadwamahmoud:
> * the migration function was not updated
> 
>     * I've pointed out some wrinkles here and there
> 
> 
> As for the migration / data trigger discrepancy, I would recommend extracting the shared functionality in a single function and then calling the same function (with different parameters) from both the trigger and the migration

I feel like that's already the case with `getCustomerStats`, I'm just using it in the migration twice to calculate the stats of both the current and following months but the main logic is in that function. Everything else is a bit too case specific to fit into one function.

packages/functions/src/migrations.ts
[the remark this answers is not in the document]
ikusteu: Right..sry, I was thrown off next/current month (but those are variable names).

ikusteu (approved):

ikusteu (approved):

PR #888, opened 2023-11-16
Normalize emails

### LINKED ISSUE #880: Make the email checks case-insensitive
There was a problem where a customer signed up with `CUSTOMER.NAME@gmail.com`, the auth entry has been created and customer self registered. Afterwards, at some point, the customer tried logging in with `customer.name@gmail.com`. Since google auth is already case insensitive, the customer was able to authenticate, but since `findCustomerByEmail` looks for an exact match, the differently cased email didn't match and the customer was prompted to self register (rather than being redirected to their `/customer_area` page)

### Tasks

- [ ] Fix the issue at hand (finding customer by email should be case insensitive)
- [ ] For all future registrations, I think it would be nice to have all emails lowercased (when signing up) and stored in the db lowercased (this is just a preference and, if the 1st task is taken care of, shouldn't have any notable effect)
- [ ] Check for other possible bugs stemming from this inconsistency and make the functionality tolerate mismatches in casing, where it makes sense

_-note: it would probably be good to create a couple of string matching utils with different levels of strictness (in terms of whitespace, casing, etc...) and use those utils for string comparisons (in existing functionality and possible future one)._

fadwamahmoud:
Fixes #880 

Normalizes emails before storage and on login, before comparison with the stored value.

Immediately after this PR is deployed to production the button `Normalize Existing Emails` needs to be pressed on the debug page.

### DIFF
--- a/packages/client/src/__tests__/cloudFunctions.test.ts
+++ b/packages/client/src/__tests__/cloudFunctions.test.ts
@@ -517,7 +517,8 @@ describe("Cloud functions", () => {
     const minimalSaul: CustomerBase = {
       name: saul.name,
       surname: saul.surname,
-      email: saul.email,
+      // non-normalized email
+      email: " SAUl@Gmail.cOm ",
       certificateExpiration: saul.certificateExpiration,
     };
     testWithEmulator(
@@ -557,6 +558,7 @@ describe("Cloud functions", () => {
           expect(Boolean(id)).toEqual(true);
           expect(customerData).toEqual({
             ...minimalSaul,
+            email: saul.email,
             secretKey,
             id,
             // Should set up empty categories
@@ -574,6 +576,8 @@ describe("Cloud functions", () => {
           expect(bookingsSnap.data()).toEqual(
             sanitizeCustomer({
               ...minimalSaul,
+              email: saul.email,
+
               secretKey,
               id,
             } as Customer)
--- a/packages/client/src/__tests__/cloudFunctions.test.ts
+++ b/packages/client/src/__tests__/cloudFunctions.test.ts
@@ -518,7 +518,10 @@ describe("Cloud functions", () => {
       name: saul.name,
       surname: saul.surname,
       // non-normalized email
-      email: " SAUl@Gmail.cOm ",
+      email:
+        " " +
+        saul.email!.replace(/.*@|o|g/g, (match) => match.toUpperCase()) +
+        " ",
       certificateExpiration: saul.certificateExpiration,
     };
     testWithEmulator(
--- a/packages/functions/src/migrations.ts
+++ b/packages/functions/src/migrations.ts
@@ -234,7 +234,7 @@ export const normalizeExistingEmails = functions
         if (!data.email) return;
 
         const { email } = data;
-        const emailNormalized = normalizeEmail(data || "");
+        const emailNormalized = normalizeEmail(email || "");
         if (email !== emailNormalized) {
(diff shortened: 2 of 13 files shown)


packages/client/src/__tests__/cloudFunctions.test.ts
```diff
@@ -517,7 +517,8 @@ describe("Cloud functions", () => {
     const minimalSaul: CustomerBase = {
       name: saul.name,
       surname: saul.surname,
-      email: saul.email,
+      // non-normalized email
+      email: " SAUl@Gmail.cOm ",
```
silviot, line 521: Nitpick: it's nice to have a single source of truth. We could use `saul.email.replace(/.*@|o|g/g, (match) => match.toUpperCase());` here and get the same result. If the email changes the part before the `@` sign will still match, and will still be uppercase. Disclaimer: [chatgpt wrote](https://chat.openai.com/share/3825208f-3a18-40a6-8ecd-8b39bc98f625) a first incarnation of that code.

packages/functions/src/migrations.ts
```diff
@@ -211,3 +212,39 @@ export const clearDeletedCustomersRegistrationAndCategories = functions
...
+  .region(__functionsZone__)
+  .https.onCall(
+    async ({ organization }: { organization: string }, { auth }) => {
+      if (!(await checkUser(organization, auth))) throwUnauth();
+
+      const db = admin.firestore();
+
+      const batch = db.batch();
+
+      const customres = await db
```
silviot, line 226: I prefer the British spelling `customers`.

packages/functions/src/migrations.ts
```diff
@@ -211,3 +212,39 @@ export const clearDeletedCustomersRegistrationAndCategories = functions
...
+        .collection(Collection.Organizations)
+        .doc(organization)
+        .collection(OrgSubCollection.Customers)
+        .get();
+
+      customres.forEach((customer) => {
+        const data = customer.data() as CustomerFull;
+        if (!data.email) return;
```
silviot, line 234: Let's try to avoid any `set` invocation unless it's really strictly needed.

I propose this change:

```
        const { email } = data;
        const emailNormalized = normalizeEmail(data || '');
        if (email !== emailNormalized) {
          batch.set(
            customer.ref,
            { email: emailNormalized },
            { merge: true }
          );
        }
```

packages/shared/src/utils/text.ts
```diff
@@ -51,3 +51,11 @@ export const checkExpected = (input: string, expected: string) => {
   const sanitizedExpected = expected.replace(/[\s,;:!?'.-]/g, "").toLowerCase();
   return sanitizedInput === sanitizedExpected;
 };
+
+/**
+ * A helper function that removes uppercase and trailing spaces from emails
+ * ie: anExampleEmail@eXAMPLE.coM into anexampleemail@example.com
+ */
+export const normalizeEmail = (email: string) => {
```
silviot, line 59: I love how there's only one `normalizeEmail`, and three places where it's used consistently :heart:

packages/functions/src/migrations.ts
```diff
@@ -153,13 +154,13 @@ export const removeInvalidCustomerPhones = functions
 
       const batch = db.batch();
 
-      const custoemrs = await db
+      const customres = await db
```
ikusteu, line 157: Good fix 😛

ikusteu (approved):

PR #891, opened 2023-11-24
Feature/bookings sanity check

fadwamahmoud:


packages/functions/src/checks/slotRelatedDocs.ts
```diff
@@ -97,6 +102,164 @@ export const findSlotAttendanceMismatches = async (
...
+
+  { id, entries }: BookingsUnpairedCheckPayload
+): Record<string, BookingsUnpairedCheckPayload> => {
+  if (entries.some((entry) => !entry.exists)) return rec;
+
+  // array length should never be not 2
+  // first element is the slot
+  if (entries.length !== 2 || entries[0].date === entries[1].date) {
+    return rec;
+  }
```
ikusteu, line 235: This is implied, no need for additional check

packages/shared/src/types/firestore.ts
```diff
@@ -425,6 +444,16 @@ export interface SlotSanityCheckReport {
...
+export interface BookingsSanityCheckReport {
+  /** ISO timestamp of the sanity check run */
+  id: string;
+  /** A record of "unpaired" bookedSlots which have no equivalent slot with the same id (keyed by slot id)*/
+  missingSlots: Record<string, BookingsUnpairedCheckPayload>;
+  /** A record of bookedSlots for which the date is mismatched across collections (keyed by slot id) */
+  invalidDateBookings: Record<string, BookingsUnpairedCheckPayload>;
+  /** A record of bookedSlots for which the interval doesn't exist in respective slot (keyed by slot id) */
+  invalidIntervalBookings: Record<string, BookingsUnpairedCheckPayload>;
+}
```
ikusteu, line 456: This is more less as I've envisioned it, only:
- I would rename `missingSlots` -> `strayBookings`
- I would rename `invalidDateBookings` -> `dateMismatches` and have each entry be `DateMismatchDoc` (conveying just enough information)
- for `invalidIntervalBookings` I would have each entry be:
```ts
{ 
  slotIntervals: Pick<SlotInterface, "intervals">,
  bookedInterval: Pick<CustomerBooking, "interval">
}
```

ikusteu (changes requested):
I left a couple of comments for smaller stuff. I would have done things differently, but let's go from the top view and make sure this works, if it does, let's move on to the autofix and then we can come back and refactor (if necessary)

### To test this out:
- start the emulators
- from `packages/client` run `rushx start:igorice`
- when prompted with confirmation, confirm (if will only override the local firestore, no fear)
- wait for the data triggers to run (you can see them running in the terminal where you're running the emulators)
- go to emulator dashboard (`localhost:4000/firesore`) and, under `igorice.web.app`, manually add `test@eisbuk.it` as admin
- on `localhost:3000` log in as `test@eisbuk.it`, navigate to debug page and run the check

Try and break stuff in emulator dashboard (in firestore) and observe if you get expected results when running the check

ikusteu:
To test these out:
- start the emulators
- from `packages/client` run `rushx start:igorice`
- when prompted with confirmation, confirm (if will only override the local firestore, no fear)
- wait for the data triggers to run (you can see them running in the terminal where you're running the emulators)
- go to emulator dashboard (`localhost:4000/firesore`) and, under `igorice.web.app`, manually add `test@eisbuk.it` as admin
- on `localhost:3000` log in as `test@eisbuk.it`, navigate to debug page and run the check

Try and break stuff in emulator dashboard (in firestore) and observe if you get expected results when running the check

packages/functions/src/checks/slotRelatedDocs.ts
```diff
@@ -178,12 +178,13 @@ export const findSlotBookingsMismatches = async (
       };
     }),
   }));
-  const strayBookings = normalisedEntries._reduce(collectStrayBookings, {});
+  const strayBookings = normalisedEntries
+    .filter((entry) => Boolean(Object.values(entry).length))
```
silviot: Shouldn't the filtering happen after the transformation from `normalizedEntries` to stray bookings (operated by `collectStrayBookings`)? I haven't tried this code, so I might be wrong, but my impression is that there will be no empty object in `normalizedEntries`, so the filtering will be a no-op.
All objects in `normalizedEntries` **MUST** include an `id` AFAICT.

fadwamahmoud: You're right. I think I forgot to remove these at some point but the desired effect is achieved by the conditions ` if (Object.keys().length > 0)` inside `collectStrayBookings` and `collectInvalidIntervalBookings`.

ikusteu:
Rightt...good catch for the "non existing slot" (2023-12-01-10 comes before 2023-12-01-9 lexicographically, I was looking in the wrong place), **but**...I think it didn't really work  with the first deletion and I'll look further into it.

As for the first point, the interval mismatch - the fact is: The interval in the slot and the interval in the booking don't match. That should have been reported, but wasn't, so that's definitely a bug.

ikusteu (approved):

PR #901, opened 2023-12-18
Feature/slot capacity

ikusteu:
Tackles the slot capacity (and private lesson functionality). I've played around with the UI and have thought of data flows, but have decided to go with the simplest working solution first and discuss further tweaks before proceeding.
Some things to discuss before continuing:
- should `private-lessons` be a regular category, unrelated to the capacity so that a combination of both is manually assigned and 1-2 people capacity + `private-lesson` category implicitly stand for private lesson slot, or should we relate the two somehow and make it so that: there's an additional checkbox `private-lesson` which automatically assigns the `private-lessons` category to the slot as well as capacity of one and locks categories/capacity. **The former is a bit ambiguous (and currently existing) solution requiring the admin to carefully set up the result, whereas the latter is a bit opinionated/limiting and the question is do we want to assume the preferences and limit the behaviour
- each slot has multiple intervals - current implementation assumes that the same capacity is applied to the entire slot, regardless of which interval was booked - this makes sense as intervals are probably just different durations of training, whereas a slot is a single training session, but I'd like to verify my reasoning on the matter
- how should the `private-lessons` category work with respect to some automatic behaviour (when off-ice is selected all of the categories are automatically selected):
  - will there be private off-ice lessons?
  - a potentially weird behaviour might arise from the fact that assigning a `private-lessons` and some other category (say `course-adults`) would render the slot bookable for an athlete that doesn't have the `private-lessons` access, but has `course-adults` category (in this case) as athlete eligibility to book a slot are determined by having at least one category from the list

### TODO
Other than the open questions above, we should also:
- update e2e tests to test the new behaviour (unit tests are there and updated), but we should also extend the e2e after the behaviour is fully defined (with respect to the questions above)


packages/client/src/store/selectors/bookings/slots.ts
```diff
@@ -118,15 +92,28 @@ export const getSlotsForCustomer = (state: LocalStore): SlotsByDay => {
...
-    const [filteredSlotsDay, isFilteredDayEmpty] = filterSlotsByCategory(
-      slotsForAMonth[date],
-      categories
-    );
-
-    // Add date to the acc object only if date not empty
-    return !isFilteredDayEmpty ? { ...acc, [date]: filteredSlotsDay } : acc;
-  }, {} as SlotsByDay);
+  //
+  // Start: Interable of [date, SlotsById (record of slots keyed by slotId)] tuples
```
silviot, line 96: I guess this is a typo: `Interable` → `Iterable`

packages/ui/src/Forms/SlotForm/SlotForm.tsx
```diff
@@ -177,27 +177,38 @@ const SlotForm: React.FC<SlotFormProps> = ({
...
+                      />
+                    ))}
+                  </fieldset>
+                  <p className="mt-2 text-sm min-h-[20px] text-red-600">
+                    {errors.type && touched.type && t(errors.type)}
+                  </p>
+                </div>
+
+                <FormField
+                  name="capacity"
```
silviot, line 206: I think this would look better with `width={FormFieldWitdh.Full}` so that it's 6 columns wide (and is the same as the type field).

fadwamahmoud (approved):

fadwamahmoud:
LGTM! I clicked around and I like that the current implementation leaves things customizable ie a slot could have a capacity of 1 and anyone could book it vs restricting access to a specific athlete. This will probably limit their requests for changes later. My only comment is negative capacity but it doesn't really affect anything.

ikusteu:
Stale, see #903

PR #905, opened 2023-12-20
Feature/trainer notes regression tests

ikusteu:
Here's my over engineered response to the PR. It includes the tests that will fail with current implementation

### DIFF
--- a/packages/client/src/__tests__/dataTriggers.test.ts
+++ b/packages/client/src/__tests__/dataTriggers.test.ts
@@ -14,6 +14,8 @@ import {
   sanitizeCustomer,
   defaultEmailTemplates as emailTemplates,
   defaultSMSTemplates as smsTemplates,
+  CustomerBookingEntry,
+  CustomerAttendance,
 } from "@eisbuk/shared";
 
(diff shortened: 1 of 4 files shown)


packages/client/src/__tests__/dataTriggers.test.ts
```diff
@@ -51,106 +95,341 @@ describe("Cloud functions -> Data triggers ->", () => {
...
+        initialAttendance: {
+          bookedInterval: Object.keys(baseSlot.intervals)[0],
+          attendedInterval: Object.keys(baseSlot.intervals)[0],
+        },
+        update: null, // null = delete the slot!
+        wantAttendance: null,
+      },
+    ]);
+
+    // test.skip("create booking: interval with notes", async () => {
```
silviot, line 300: I'm curious: why commenting out a skipped test? I'd say it should be either be left alone or removed, but not commented out. I'm probably missing a possible reason to do this.

ikusteu: Thanks, this is a (forgotten) leftover. Will fix with in the currently open PR when resolving conflicts

PR #906, opened 2023-12-21
Feature/monthly attendance tweaks

### LINKED ISSUE #896: Monthly attendance tweaks
Currently the monthly attendance has four rows per customer. They're needed to communicate "Ice booked hours", "ice attended hours", "off-ice booked hours" and "off ice attended hours".

It's preferrable to hide the specific number of attended hours (it could still be shown on hover) and use color coding to communicate alignment between booked and attended: when they don't match the cell can have a red or yellow background.

This way the table only needs two rows per athlete.

Also omit off-ice rows for athlete that didn't book any.

We understand and accept that the table will look uglier after this change.

fadwamahmoud:
Fixes #896

### DIFF
--- a/packages/ui/src/Table/Table.tsx
+++ b/packages/ui/src/Table/Table.tsx
@@ -2,7 +2,10 @@ import React from "react";
 
 import TableCell, { CellType } from "./TableCell";
 
-type TableItem = Record<string, string | number | boolean | null>;
+type TableItem = Record<
+  string,
+  string | number | boolean | null | { booked: number; delta: number }
+>;
 type TableHeaders<T> = Record<keyof T, string>;
 
 interface RenderRow<T extends TableItem = TableItem> {
@@ -37,6 +40,8 @@ const Table = <T extends TableItem>({
     <tbody className="bg-gray-50">{items.map(renderRow)}</tbody>
   </table>
 );
+// item :tableItem
+// itemArr :tableItem[]
 
 const defaultRenderRow: RenderRow = (item, rowIx, itemArr) => {
   const bgClasses = rowIx % 2 === 0 ? undefined : "bg-gray-50";
--- a/packages/ui/src/Table/Table.tsx
+++ b/packages/ui/src/Table/Table.tsx
@@ -40,8 +40,6 @@ const Table = <T extends TableItem>({
     <tbody className="bg-gray-50">{items.map(renderRow)}</tbody>
   </table>
 );
-// item :tableItem
-// itemArr :tableItem[]
 
 const defaultRenderRow: RenderRow = (item, rowIx, itemArr) => {
   const bgClasses = rowIx % 2 === 0 ? undefined : "bg-gray-50";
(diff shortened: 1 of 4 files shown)


silviot (changes requested):
I think it should display the hours booked, and on hover the difference (if any) with the hours attended.

Right now it shows the difference with the hours attended, and on hover the hours booked (IIUC).

They should be swapped.

silviot (approved):

packages/ui/src/Table/Table.tsx
```diff
@@ -2,7 +2,10 @@ import React from "react";
 
 import TableCell, { CellType } from "./TableCell";
 
-type TableItem = Record<string, string | number | boolean | null>;
+type TableItem = Record<
+  string,
+  string | number | boolean | null | { booked: number; delta: number }
+>;
```
ikusteu, line 8: Normally, I would advise against this: By allowing `{ booked: number; delta: number }` as a value you're turning the `Table` (imagined as a generic component) in to an opinionated one.

Buuut, in the spirit of pragmatism, **this table is used only once** so there's no harm here - _If we had more time to devote to this, I would even advise merging the two components into AttendanceVarianceTable, but this is ok._

ikusteu (approved):
All in all, this seems to do what it's supposed to. I find it counterproductive for me to spend some significant amount of time going through the code. Here are some soft notes (you can chose to fix them up or not, it's up to you, all in all I wouldn't spend too much time on them, especially if the customer is happy):
- I'm a bit confused with the colour coding - I understand green is where the booked/attended match and red is where they don't (but only on ice slots), whereas yellow is where they don't match on off ice...but: **if it's clear for the user, I'm good**
- hover: I would make the entire table have `select-none` class and would make it so that diff appears when you hover over the table cell (not the text inside the cell)
- diff: Having `1` for "attended 1 hour more than booked" and `-1` for one hour less than booked is mathematically correct, but in terms of clearer communication I would include the `+` sign:
  - `+1` attended 1 hour more
  - `-1` attended 1 hour less

All in all, this can go in AFAIC

fadwamahmoud:
> All in all, this seems to do what it's supposed to. I find it counterproductive for me to spend some significant amount of time going through the code. Here are some soft notes (you can chose to fix them up or not, it's up to you, all in all I wouldn't spend too much time on them, especially if the customer is happy):
> 
>     * I'm a bit confused with the colour coding - I understand green is where the booked/attended match and red is where they don't (but only on ice slots), whereas yellow is where they don't match on off ice...but: **if it's clear for the user, I'm good**
> 
>     * hover: I would make the entire table have `select-none` class and would make it so that diff appears when you hover over the table cell (not the text inside the cell)
> 
>     * diff: Having `1` for "attended 1 hour more than booked" and `-1` for one hour less than booked is mathematically correct, but in terms of clearer communication I would include the `+` sign:
>       
>       * `+1` attended 1 hour more
>       * `-1` attended 1 hour less
> 
>     * Is the colour coding of the rows (blue/yellow) enough for the admin to distinguish between ice and off ice? (I'm guessing it is and this was at their request)
> 
> 
> All in all, this can go in AFAIC

I think your comments are valid and yes the color code made me itch a bit but Silvio approved it and they don't seem to be concerned about much but the table being compact so I'd rather not work on this any longer until they use it and wait see if they want things differently.

packages/ui/src/Table/Table.tsx
fadwamahmoud: I knew this would give you the ick...but yeah I was thinking the same thing

PR #908, opened 2023-12-24
Feature/athlete avatar menu

ikusteu:
Creates a dropdown menu on user avatar in `/customer_area`. The menu will be used to switch between multiple accounts (if that's the case). For now, only the logout functionality is wired in, but the component supports full functionality which will be wired in in the later PRs

This is the UI  side of things in the single auth / multiple athletes functionality. I imagine this being a storybook + click through review (but please review the code as well...).

### DIFF
--- a/packages/ui/src/AthleteAvatarMenu/AthleteAvatarMenu.tsx
+++ b/packages/ui/src/AthleteAvatarMenu/AthleteAvatarMenu.tsx
@@ -1,0 +1,73 @@
+import React from "react";
+
+import { Customer } from "@eisbuk/shared";
+import { Plus, PowerCircle } from "@eisbuk/svg";
+
+import { shortName } from "../utils/helpers";
+
+export interface AthleteAvatarMenuProps {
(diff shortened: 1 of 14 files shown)


fadwamahmoud (approved):

packages/ui/src/AthleteAvatarMenu/AthleteAvatarMenu.tsx
```diff
@@ -0,0 +1,73 @@
...
+      ) : null}
+
+      <div className="mt-4">
+        <p className="text-xs text-gray-600 uppercase">Actions</p>
+        <div className="pl-3 pt-2">
+          <button onClick={onAddAccount} className={itemClasses}>
+            <span className="align-middle inline-block h-5 w-5 mr-1">
+              <Plus />
+            </span>
+            <span className="align-middle">Add account</span>
```
silviot, line 57: This one is not yet working. It's better to hide it for now.

PR #910, opened 2024-01-02
Send bulk booking emails

### LINKED ISSUE #898: Bulk send email with booking link
Trainers need a way to remind athletes to make their bookings once a month (most likely twice - five days before expiration and one day before).

Admins might have a screen to do the bulk sending, with a prepopulated template that they can edit.

fadwamahmoud:
Fixes #898


packages/client/src/features/modal/components/SendBulkBookingsLinkDialog/SendBulkBookingsLinkDialog.tsx
```diff
@@ -0,0 +1,102 @@
...
+} from "../SendBookingsLinkDialog/utils";
+
+import { updateOrganizationEmailTemplates } from "@/store/actions/organizationOperations";
+import { getMonthDeadline } from "@/store/selectors/bookings";
+
+type SendBulkBookingsLinkProps = BaseModalProps & { customers: Customer[] } & {
+  method: ClientMessageMethod;
+  emailTemplates: OrganizationData["emailTemplates"];
+  actions: FormikHelpers<OrganizationData["emailTemplates"]>;
+  calendarDay: DateTime;
```
ikusteu, line 28: We never want to store functions in Redux. Redux doesn't play well with non-serialisable objects (e.g. functions)

packages/client/src/utils/helpers.ts
```diff
@@ -100,3 +102,74 @@ export const replaceHTMLTags = (template: string) => {
...
+ * @returns A function that takes a button value and inserts it into the input field.
+ */
+
+export const insertValuePlaceholder =
+  (
+    setFieldValue: (
+      field: string,
+      value: any,
+      shouldValidate?: boolean | undefined
+    ) => Promise<void | FormikErrors<OrganizationData>>,
```
ikusteu, line 120: Since you know the shape of the form, you could do something like this: 
```ts
const insertValuePlaceholder = <F extends keyof OrganizationData>(
  field: F, // Constraining field to keys of OrganizationData
  value: OrganizationData[F], // Not using 'any' 
   ...) => Promise<void | FormikErrors<OrganizationData>>
```

ikusteu (changes requested):
1. When open, the athlete list is overflowing the screen:
Screenshot 2024-01-22 at 07 47 07

This might be rather hard to fix, therefore, I would suggest creating a box in the html flow (no `absolute`) so that the box is bound by other elements (it's not floating above) and is scrollable within (in case the number of athletes is such that the content is larger than the box. It doesn't have to be huge: I think fitting 5-6 athletes and showing that there are more to scroll should do it.

2. I don't think 'Save' is the right label for the button (probably 'Send')
Screenshot 2024-01-22 at 07 47 29

packages/ui/src/CheckboxCustomerList/CheckboxCustomerList.tsx
```diff
@@ -0,0 +1,94 @@
...
+                </div>
+
+                <div className="flex items-center justify-end text-gray-500">
+                  <div className="flex items-center w-16">
+                    <div className="h-4 ">
+                      <Ice />
+                    </div>
+
+                    {customer.bookingStats
+                      ? `${customer.bookingStats[monthStr].ice} hrs`
```
silviot, line 73: I was trying this on https://igoriceteam--feature-bulk-emails-j0ed7n0i.web.app/booking_emails
and I had this problem:
https://sentry.libroc.co/organizations/eisbuk/issues/166/?query=is%3Aunresolved

```
Cannot read properties of undefined (reading 'ice')
```
I think we should be defensive here and cope with a possibly missing `customer.bookingStats[monthStr]`.

packages/ui/src/CheckboxCustomerList/CheckboxCustomerList.tsx
```diff
@@ -0,0 +1,94 @@
...
+                      : ` - `}
+                  </div>
+                  <div className="flex items-center w-16">
+                    <div className="h-4">
+                      <OffIce />
+                    </div>
+                    {customer.bookingStats
+                      ? `${customer.bookingStats[monthStr]["off-ice"]} hrs`
```
silviot, line 81: I guess the same thing I mentioned above applies here as well.

packages/client/src/pages/booking_emails/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+                    type="submit"
+                  >
+                    {t(ActionButton.Save)}
+                  </Button>
+                </div>
+              }
+            >
+              <EmailTemplateSettings
+                onCheckboxChange={handleCheckboxChange}
+                customers={allCustomers}
```
silviot, line 167: This should be `nonDeletedCustomers`, not `allCustomers`. I.e. we need to filter out deleted customers.

packages/ui/src/CheckboxCustomerList/CheckboxCustomerList.tsx
fadwamahmoud: Shit. Yeah, I had actually fixed this then lost it and forgot about it amongst the purge/reclone mess.

ikusteu (approved):
Good job on the overflow fix 👌

silviot (changes requested):
Before this goes in two things are needed:

* show next month's stats, not this one's
* when the user selection dropdown is closed there's currently no indication of the selected recipients. They should be visible when clicking "submit"

silviot (approved):

PR #916, opened 2024-01-17
Update/simplify sorting

ikusteu:
A quick refactor to keep compare (sort callback) functions in the shared package (single location) for reusability with the current codebase, as well as upcoming e2e tests


ikusteu:
Merging this in as it has been thoroughly tested and includes no breaking changes, but blocks the following steps.

PR #917, opened 2024-01-17
Feature/single account multiple athletes

### LINKED ISSUE #904: Allow for single auth user to manage multiple athletes
This is a quick recap, from memory, on our talk on the matter.

Sometimes a single auth would manage multiple athlete's accounts (e.g. one parent managing bookings for their children). Currently we don't support that and the behaviour is more/less undefined (depends on how they go about it).

The proposed updates:
- [ ] When querying athlete by email/phone, the function should return all matches (instead of the first match, as is currently the case)
- [ ] On login, single user - single athlete would take to the athlete's page (as is the current implementation)
- [ ] On login, single user - multiple athletes, the user is redirected to a new page (a prompt really) prompting them to chose the athlete they want to act for
- [ ] There should be a user dropdown (on click of user avatar) with:
  - currently logged in athlete (emphasised: bold or smtn)
  - a list of alternative athletes (not currently active, but also managed by the same account), empty if one - one user - athlete
  - add new athlete -> allowing the user to add additional athlete for management
  - logout button
- [ ] When adding of new athlete, the flow should be the same as for onboarding (self-register)

### Enhancements (TBD stuff):
- how should deleting of athlete account work
- how should email notifications work (probably the same way they already do since each athlete, not user, would have email/phone assigned to them)
- should updating of one athlete's email/phone be reflected with the other athletes managed by the same user ??

ikusteu:
Fixes #904


silviot (changes requested):
Looks great in general. Needs a fix though.

I tested this in production (AFAIU there are no changes to cloud functions in this PR, so it's safe to test it through the preview channel deployed by the CI) and I was able to:
* register a new athlete
* see the newly registered athlete amongst the choices in `/select_account`

I was not able to:

* see my other, already existing, account with the same email
* see my other, already existing, account with the same phone number

packages/ui/src/AthleteAvatarMenu/AthleteAvatarMenu.tsx
```diff
@@ -50,12 +50,12 @@ const AthleteAvatarMenu: React.FC<AthleteAvatarMenuProps> = ({
       <div className="mt-4">
         <p className="text-xs text-gray-600 uppercase">Actions</p>
         <div className="pl-3 pt-2">
-          {/* <button onClick={onAddAccount} className={itemClasses}>
+          <button onClick={onAddAccount} className={itemClasses}>
             <span className="align-middle inline-block h-5 w-5 mr-1">
               <Plus />
             </span>
             <span className="align-middle">Add account</span>
```
silviot, line 57: `Add account`, `Other accounts` and `actions` need to be translated. I propose to remove `Other accounts` and `Actions`, and replace the latter with a line separator.

ikusteu:
> Looks great in general. Needs a fix though.
> 
> I tested this in production (AFAIU there are no changes to cloud functions in this PR, so it's safe to test it through the preview channel deployed by the CI) and I was able to:
> 
> * register a new athlete
> * see the newly registered athlete amongst the choices in `/select_account`
> 
> I was not able to:
> 
> * see my other, already existing, account with the same email
> * see my other, already existing, account with the same phone number

I've used eisbuk.web.app (the preview channel) to test this out: I've created a new athlete (with my existing account). I was able to see both and navigate between...

Did you (maybe) mean: You had two accounts with the same email and expected to see them both in the selection (now with the multiple athlete update)...if so, yeah, that should be working, but if it's not, maybe we should debug that together...

packages/client/src/pages/select_account/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+import ErrorBoundary from "@/components/atoms/ErrorBoundary";
+
+import { getOrganization } from "@/lib/getters";
+
+import { getAllSecretKeys } from "@/store/selectors/auth";
+import { getOtherBookingsAccounts } from "@/store/selectors/bookings";
+
+const SelectAccount: React.FC = () => {
+  const secretKeys = useSelector(getAllSecretKeys) || [];
+  const accounts = useSelector(getOtherBookingsAccounts(""));
```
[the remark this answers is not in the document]
ikusteu: Ok, I did this, but it should actually never happen:
- customers with fewer than 2 secret keys wouldn't (at least shouldn't) ever reach the page
- there should be the same number of accounts as secret keys

In case this redirect happens, that indicates there's something broken in the code or the db. It could happen, but the question is: if we're watching out for that (unlikely) scenario, should we also add some way of reporting the mishap to Sentry 🤷

packages/ui/src/AthleteAvatarMenu/AthleteAvatarMenu.tsx
ikusteu: Here, which one do you prefer:

The old one:
Screenshot 2024-01-26 at 07 54 47

The suggested update:
Screenshot 2024-01-26 at 08 00 10

I favour the old one slightly, but I think both are good...

ikusteu: The new one does look nicer, but the old one is more telling...

silviot: I like the new one: there are also less strings to translate (and I don't dislike its appereance).

silviot (approved):

PR #926, opened 2024-01-23
Fix some problems around email sending:

### LINKED ISSUE #919: Email delivery problem
image

The email delivery function fails with this message:

> Error at field 'to': "must be a valid email string"

ikusteu:
* Create 'enqueueEmailForDelivery' util and use it in all cloud functions attempting to send emails
* Update the email string validations to match a wider range of (supported) email strings

Fixes #919 

Problem:
- athleteSelfRegister cloud function added emails to the queue directly, without some validations (that, naturally, introduced an error: email payload should be nested under `payload` property, and in this case it was spread on the top level of the object) - Fixed by the 1st point ☝️ 
- the email string (for one of the failed emails) included a `+` sign, which we didn't support as valid email string - Fixed by the 2nd point ☝️


common/scripts/emulators-ci
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+repo_root_dir=$(CDPATH= cd -- "$(dirname -- "$0")"/../.. && pwd)
+
+trap "kill 0" SIGINT
+
+(cd $repo_root_dir/packages/functions && rushx build ) &
+export CI=true && firebase emulators:start --project eisbuk $@
```
silviot, line 8: I have a few reservation with this approach:

* there is no explanation in this commit about why this has been done; stating that it is "to fix ci tests" is not enough: why are they broken? Why should this fix them? What other possible solutions were evaluated and discarded?
* The `CI` variable [will already have a value](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) when run on github actions. So the utility of this script is to pretend to be in the CI when running locally. There is no explanation about why this would be a good idea.

packages/functions/src/sendSMS/deliver.ts
```diff
@@ -106,3 +107,10 @@ export const deliverSMS = functions
       })
     )
   );
+
+export const deliverSMSTesting = functions
+  .region(__functionsZone__)
+  .firestore.document(
+    `${Collection.DeliveryQueues}/{organization}/${DeliveryQueue.EmailQueue}/{emailDoc}`
+  )
+  .onWrite(skippedDataTrigger("deliverEmailTesting", (data) => data?.payload));
```
silviot, line 116: I don't understand the purpose of this trigger.

silviot: Ok, I needed to read the code below. I now understand it. Sorry for the noise.

ikusteu: Yeah...the main purpose: have a conditional export (have the export in `index.ts` have something to fall back to)...thinking about it now, maybe I could have used `null` as a fallback...

silviot: On third thought, I believe this is not necessary. I'll open a PR to propose [these changes](https://github.com/eisbuk/EisBuk/commit/a1046df74a84461e1a8106268e001cab013943a7).

silviot: @ikusteu Sorry for ignoring your comment. Github did not update the UI for me, so I missed it.

silviot (approved):

PR #932, opened 2024-01-27
Fix/query auth status

### LINKED ISSUE #930: `queryAuthStatus` and `queryAuthStatus2` are leaking user secret keys
Given a user email or phone, an anonymous user can query these functions to get their secret key(s). The functions should check that the `authString` provided in the request matches the currently logged in used, and return `{"result":{"isAdmin":false,"secretKeys":[]}}` if they don't.

`authString` needs to be removed from `QueryAuthStatusPayload`. The authentication string needs to be pulled from the functions context, [as documented here](https://firebase.google.com/docs/functions/callable?gen=1st) in a similar way to [what we already do in other cloud functions](https://github.com/eisbuk/EisBuk/blob/0182d86e4d57f7ea2dc44547ec7d70fe1c638894/packages/functions/src/checks/https.ts#L30).

ikusteu:
Fixes #930

### DIFF
--- a/packages/client/src/__tests__/auth.test.ts
+++ b/packages/client/src/__tests__/auth.test.ts
@@ -2,17 +2,17 @@
  * @vitest-environment node
  */
 
-import { describe, expect } from "vitest";
+import { afterEach, describe, expect } from "vitest";
 import {
   httpsCallable,
   HttpsCallableResult,
   FunctionsError,
 } from "@firebase/functions";
 
-import { Collection, AuthStatus, HTTPSErrors } from "@eisbuk/shared";
+import { AuthStatus, HTTPSErrors } from "@eisbuk/shared";
 import { CloudFunction } from "@eisbuk/shared/ui";
 
-import { adminDb, functions } from "@/__testSetup__/firestoreSetup";
+import { adminDb, auth, functions } from "@/__testSetup__/firestoreSetup";
 
 import { getCustomerDocPath } from "@/utils/firestore";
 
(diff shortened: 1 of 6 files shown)


packages/functions/src/auth.ts
```diff
@@ -85,12 +26,21 @@ export const queryAuthStatus2 = functions
...
+      "queryAuthStatus",
+      async (
+        payload: QueryAuthStatusPayload,
+        { auth }
+      ): Promise<AuthStatus> => {
+        // Organization is required to even start the functionality
+        checkRequiredFields(payload, ["organization"]);
+        const { organization } = payload;
+
+        await checkUser(organization, auth);
```
silviot, line 38: `checkUser` has no side effects AFAICT.
This invocation can probably be removed.

packages/client/src/__tests__/auth.test.ts
```diff
@@ -2,103 +2,59 @@
...
+import { signInEmail } from "@/__testUtils__/auth";
 
 describe("Test authentication", () => {
+  afterEach(() => {
+    auth.signOut();
+  });
+
   describe("Test queryAuthStatus", () => {
-    const queryAuthStatus = (organization: string, authString: string) =>
+    const queryAuthStatus = (organization: string) =>
```
silviot, line 31: We should probably [also change the invocation](https://github.com/eisbuk/EisBuk/blob/711f4120955f0677096a8b1957f3fee3cf9f3510/packages/client/src/store/actions/authOperations.ts#L67) of this cloud function to omit `authString` for consistency.

packages/shared/src/types/cloudFunctions.ts
```diff
@@ -27,8 +27,6 @@ export type CreateAuthUserPayload = Partial<{
 export interface QueryAuthStatusPayload {
   /** Current organization */
   organization: string;
-  /** String used to authenticate user, either email or phone */
-  authString: string;
```
silviot, line 31: We should probably [also change the invocation](https://github.com/eisbuk/EisBuk/blob/711f4120955f0677096a8b1957f3fee3cf9f3510/packages/client/src/store/actions/authOperations.ts#L67) of this cloud function to omit `authString` for consistency.

silviot (approved):
I left a couple of comments about things that can be improved, but they're not really blockers. I'm opening a new ticket about using firestore indices. The small fixes can be done in that context too.

PR #934, opened 2024-01-27
Improve deleted customer page redirects:

### LINKED ISSUE #921: Make the `/deleted` route redirect ot the root if the user is not authenticated
The url with `/delete` has been erroneously shared with athletes, with instructions to use it to register. Athletes were frustrated that they could not register. It would be best if, if not authenticated, the `/deleted` route redirected to the login/register screen.

ikusteu:
* If no secret key, redirect to default page with respect to auth state
* If customer not deleted, redirect to customer area for the customer

Fixes #921


silviot (approved):
I didn't go through the code; I tested the functionality and works great.

PR #935, opened 2024-01-27
Optimise 'queryAuthStatus' by using two 'where' clauses instead of iterating over all docs

### LINKED ISSUE #933: Improve `queryAuthStatus` firestore customer query
Currently `queryAuthStatus` (a function that is called by every client on every page load, and that blocks usage until it returns) [reads in all customers](https://github.com/eisbuk/EisBuk/blob/711f4120955f0677096a8b1957f3fee3cf9f3510/packages/functions/src/auth.ts#L49) and loops through them to find matches for phone/email.

It could instead use two indices and two [`where` queries](https://firebase.google.com/docs/firestore/query-data/index-overview#single-field-indexes).

I believe [the indices are automatically created by firebase](https://firebase.google.com/docs/firestore/query-data/index-overview#automatic_indexing) already.

There were two small fixes still to do in #932 that makes sense to do in this context.

It also makes sense to fix [this sentry entry](https://sentry.libroc.co/organizations/eisbuk/issues/158/?project=5&query=is%3Aunresolved) in this context. The gist of it is that the line

```
        const { email, phone_number: phone } = auth!.token!;
```
in `auth.ts` causes an error if the function is called by an unauthenticated user. I was testing the function with `curl`, so it's very much possible a similar thing never happens using the app. Still, it makes sense to not throw an exception in that case.

ikusteu:
Fixes #933

### DIFF
--- a/packages/client/src/__tests__/auth.test.ts
+++ b/packages/client/src/__tests__/auth.test.ts
@@ -93,5 +93,16 @@ describe("Test authentication", () => {
         }
       }
     );
+
+    testWithEmulator(
+      "should not explode if user is not logged in",
+      async () => {
+        const { data } = await httpsCallable(
+          functions,
+          CloudFunction.QueryAuthStatus
+        )({ organization: "idk" });
+        expect(data).toEqual({ isAdmin: false, secretKeys: [] });
+      }
+    );
   });
 });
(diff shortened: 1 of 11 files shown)


silviot (approved):
Excellent! It can go in as is, but it would be great to also have fixes for these:

* https://github.com/eisbuk/EisBuk/pull/932#discussion_r1468448514
* https://github.com/eisbuk/EisBuk/pull/932#pullrequestreview-1847064933

silviot:
Also the check on `auth` before the line

```
        const { email, phone_number: phone } = auth!.token!;
```

that prevents [this sentry entry](https://sentry.libroc.co/organizations/eisbuk/issues/158/?project=5&query=is%3Aunresolved) is not yet in the PR.

PR #937, opened 2024-01-28
Fix/invisible buttons

### LINKED ISSUE #922: Adjust display of trash cans on slot edit screen
Some slots can be deleted, others can't. Slots with bookings can't be deleted. If you press the trash can on these, you get a message stating that the slot has bookings and can't be deleted.

But the trash cans are always visible when they're "not working" (i.e. on slots with bookings) but only display on hover when they are.

If there's a difference, it should be the other way round. Best would be to show them greyed out when not available.

The same applies to copy/paste and create slot buttons: they are invisible.
See https://storage.googleapis.com/cypress-logs-bucket/2024-01-23/7631556077/storybook-ui/index.html?path=/story/slot-card--with-action-buttons: the slot has to invisible buttons in the bottom right.

ikusteu:
Fixes #922 

☝️ the problem wasn't the reversed booked/non-booked logic, it was that the buttons where white, when they should have been gray-500, making the disabled buttons, gray-( < 500 ) seem accented (compared to white, invisible ones)

### DIFF
--- a/packages/client/src/main.css
+++ b/packages/client/src/main.css
@@ -3,7 +3,6 @@
 @import "tailwindcss/components";
 
 @import "tailwindcss/utilities";
-@import "../../ui/dist/style.css";
 
 @font-face {
   font-family: "Fredoka One";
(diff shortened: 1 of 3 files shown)


silviot (approved):
I have to confess I haven't looked thoroughly into Tailwind yet, and thus don't fully understand this PR. It seems to work though.

PR #938, opened 2024-01-28
Use 'emailNameFrom' when sending email (and use 'emailFrom' as fallback)

### LINKED ISSUE #931: "Organization Email Name" setting is not honoured
I tried sending an email, and it worked.

However the "Organization Email Name" setting had no effect.

ikusteu:
Fixes #931 

This needs to be deployed first (the updated `sendEmail`) and checked that way...I wrote a test verifying that the expected payload is handed off for delivery, but can't verify for sure that it will work e2e as expected (until and actual mail has beed sent, that is...).

### DIFF
--- a/packages/client/src/__tests__/sendEmail.test.ts
+++ b/packages/client/src/__tests__/sendEmail.test.ts
@@ -202,8 +202,7 @@ describe("SendEmail", () => {
     expect(deliveryDocData).toEqual(
       expect.objectContaining({
         payload: expect.objectContaining({
-          // Using email name from instead of email from
-          from: "Eisbuk Team",
+          from: `Eisbuk Team <${emailSetup.emailFrom}>`,
           to: saul.email,
           bcc: emailSetup.emailBcc,
         }),
--- a/packages/client/src/__tests__/sendEmail.test.ts
+++ b/packages/client/src/__tests__/sendEmail.test.ts
@@ -48,6 +48,16 @@ type SendEmailTest =
       >;
     };
 
+const orgSetup = {
+  doLogin: true,
+  setSecrets: true,
+};
+
+const emailSetup = {
+  emailFrom: "eisbuk@email.com",
+  emailBcc: "bcc@gmail.com",
+};
+
 /**
  * Runs table tests from the payload passed in (for each email type)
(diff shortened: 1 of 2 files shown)


packages/functions/src/sendEmail/https.ts
```diff
@@ -90,7 +91,7 @@ export const sendEmail = functions
 
       // Construct an email for process delivery
       const email = {
-        from: emailFrom,
+        from: emailNameFrom || emailFrom,
```
silviot, line 94: I think we should combine name and email here in the form of a "mailbox": https://superuser.com/questions/1625686/what-is-the-email-address-format-name-email-called

packages/client/src/__tests__/sendEmail.test.ts
```diff
@@ -159,4 +163,51 @@ describe("SendEmail", () =>
...
+    // Check the process document
+    const deliveryDocPath = deliveryDocumentPath;
+    const deliveryDoc = await adminDb.doc(deliveryDocPath).get();
+    const deliveryDocData = deliveryDoc.data();
+
+    expect(deliveryDocData).toEqual(
+      expect.objectContaining({
+        payload: expect.objectContaining({
+          // Using email name from instead of email from
+          from: "Eisbuk Team",
```
silviot, line 206: That's not a vali email address

silviot (changes requested):

packages/functions/src/sendEmail/https.ts
ikusteu: Oook...had the same idea, but decided to cut it as I wasn't sure if it was my overcomplicating things (plus I didn't know the standard format, so thanks for this...)

packages/client/src/__tests__/sendEmail.test.ts
ikusteu: I assume the previous comment fix will address this too, right?

silviot (approved):

packages/client/src/__tests__/sendEmail.test.ts
silviot: Yeah. I wanted to add that for clarity

PR #945, opened 2024-03-20
Fix/bookings deadline

### LINKED ISSUE #840: Improve booking extension date logic
Currently if an athlete is given a chance to book a certain month up to a late day (say the 29th) the following month results unbookable from the 5th to last day on. So today (the 27th) you can see the problem if you set a date like 29th.
It should instead also be bookable.

ikusteu:
Fixes #840


silviot:
I just tried to extend it to let an athlete book slots in March, but I was not able to.
image
image

ikusteu:
@silviot I've extended the debug box functionality and wired in the translated strings. Could you, please, take a look and provide italian translations?

_Just a reminder: we went with the existing implementation and can update (extend the allowed booking timespan if they really need it_

silviot:
image
I would expect the countdown to read `3d3h`.

image
The bookings should not be locked in this case.

image
The bookings should not be locked in this case.

ikusteu:
Gaah...on it!

ikusteu:
1. If the regular deadline hadn't passed yet, it is displayed (yellow box), and only when the regular deadline had passed, but extended date hadn't, the extended date is displayed (red box). The reasoning is:
  - I'm guessing the extended date won't be provided unless the original deadline had passed
  - let's assume it was the other way around (extended date having priority) - it would be awkward to click 'Finalise' and then have the first deadline countdown displayed

2.
3. Well...I see nothing wrong, considering the extended date had passed two years ago 😂

silviot:
@ikusteu hmmm... right. Yeah, with the correct date it works as expected. :facepalm:

silviot (approved):

PR #946, opened 2024-03-27
Debug setup 1

ikusteu:
This is draft as it's not a real PR, but rather a dummy one for easier changes + comments. I will push some debug setups and logs of the debugging process here.

### DIFF
--- a/packages/functions/src/constants.ts
+++ b/packages/functions/src/constants.ts
@@ -1,19 +1,29 @@
 import * as Sentry from "@sentry/serverless";
+import functions from "firebase-functions";
 
 export const __functionsZone__ = "europe-west6";
 export const __projectId__ = process.env.PROJECT_ID;
 export const __smsUrl__ = "https://gatewayapi.com/rest/mtsms";
 export const __isEmulator__ = process.env.FUNCTIONS_EMULATOR === "true";
 
-export const __sentryDSN__ = process.env.FUNCTIONS_SENTRY_DSN;
-export const __sentryRelease__ = process.env.REACT_APP_SENTRY_RELEASE;
-export const __enableSentry__ =
-  !process.env.FUNCTIONS_EMULATOR && Boolean(__sentryDSN__);
+// Sentry dsn is set up to send all logs to the web-sink for easier inspection
+export const __sentryDSN__ = "http://admin:password@127.0.0.1:3001/1234567"; // process.env.FUNCTIONS_SENTRY_DSN;
+export const __sentryRelease__ = "test-release"; // process.env.REACT_APP_SENTRY_RELEASE;
+export const __enableSentry__ = true;
+// !process.env.FUNCTIONS_EMULATOR && Boolean(__sentryDSN__);
+
+// (globally) Override console.log to reroute logs to Firebase Functions logger
+// In functions we already us the functons logger, but for 3rd party libs (such as Sentry client), with debug on,
+// we need to reroute console.log to functions.logger.log (else they won't be displayed anywhere)
+console.log = functions.logger.log;
+console.info = functions.logger.info;
+console.error = functions.logger.error;
 
 if (__enableSentry__) {
   Sentry.init({
     dsn: __sentryDSN__,
     release: __sentryRelease__,
     tracesSampleRate: 1.0,
+    debug: true,
   });
 }
--- a/packages/functions/src/constants.ts
+++ b/packages/functions/src/constants.ts
@@ -1,5 +1,5 @@
 import * as Sentry from "@sentry/serverless";
-import functions from "firebase-functions";
+// import functions from "firebase-functions";
 
 export const __functionsZone__ = "europe-west6";
 export const __projectId__ = process.env.PROJECT_ID;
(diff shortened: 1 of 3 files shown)


ikusteu:
Before the debug findings, a little context:
I've set up a simple server program, listening to `localhost:3001` and logging every request out + responding with preset status code (default: 200, but can be changed between requests), [more here](https://github.com/ikusteu/web-sink)

_note: I wrote it in Go as I've never learned to use Express.js and didn't want to bother with Node.js' (boilerplaty) native http handling. However, the server is rather simple + I've commented the code through and through for it to be understandable to anybody not familiar with Go's std lib._

Furthermore, I've set up Sentry client (in Eisbuk app) to use `localhost:3001` as `__sentryDsn__` as it's easier to inspect requests sent out + using the web-sink, we can simulate different response codes from "sentry" service.

ikusteu (commented):
## 1st round of debugging: http request using curl

- added logging every step of the way (to verify the functions are being wrapped in sentry handlers)
- used `testException` http function and observed the behaviour (logs + web-sink)
- tested calling the function with:
  - error = true web-sink on and responding with `200`
  - error != true (`textException` not throwing), web-sink on and responding with `403`
  - error != true web-sink off (nothing listening to `:3001`)
  - web-sink listening (res code not important) and logging + throwing errors within sentry function wrappers (to test where the code communicates with Sentry)

### Error == true, web sink on and responding with 200

Called `testException` using curl, the exception was caught by Sentry client and sent to `localhost:3001` (as expected). Web-sink logged:
Screenshot 2024-03-27 at 12 21 32
...and so on
Function logs (confirming the error was caught by Sentry wrapper):
Screenshot 2024-03-27 at 12 23 49

### Error != true, web sink on and responding with 403

Called `testException` using curl, the function passed without exception and sent the report to `localhost:3001`.

 Web-sink logged:
Screenshot 2024-03-27 at 12 25 41

A significantly shorter string (no stack traces, compared to the caught and reported exception)

The functions did log the not-so-happy response (403), but that didn't crash the function and the result was ok.

Functions logs:
Screenshot 2024-03-27 at 13 13 59

Curl result:
Screenshot 2024-03-27 at 13 14 54

### Error != true web-sink off (nothing listening to `:3001`)

Used curl to ping the `testException` with exception != true. The function ran before any error was thrown (see function logs), error was (in this case) actually thrown (but only after the main part of the function ran) and the function gave a successful response (see curl res).

Functions logs:

Screenshot 2024-03-27 at 13 37 36


Curl res:

Screenshot 2024-03-27 at 13 39 28

### Error thrown in function wrapper (after the function ran)

I've added `throw new Error("here")` right before Sentry client's `flush` and `curl`ed the function with error != true. Web sink didn't receive anything. Did it again with error thrown right after `flush` and verified that `flush` is, in fact, the first point of contact between (wrapped) functions and "Sentry" (`localhost:3001`), as web-sink did receive the payload, in this case. 

Code with error before flush (added to both happy path an error as the former error gets caught with the `catch` block):
Screenshot 2024-03-27 at 13 48 05

Code with error after flush is the same (only the error is a row below).

Function logs were the same in both cases:

Screenshot 2024-03-27 at 13 50 24

Interesting thing is that curl got the pong result both times (indicating that sentry wrapper error didn't affect the response in any way):
Screenshot 2024-03-27 at 13 54 16

ikusteu:
## Step 2

I've created a data trigger (`dataTriggerWithFailingSentry`) to test out whether or not failing sentry communication will result in data trigger not being executed / being rolled back at error.

I've tested this with:
- web-sink on, listening to `localhost:3001` and responding with `200`
- web-sink on, listening to `localhost:3001` and responding with `403`
- web-sink off, simulating the server being down

### Web-sink on, res: 200

Nothing to report here: the trigger ran as expected, the new document had been created and logs + web-sink output are consistent with expectations

### Web-sink on, res: 403

The trigger ran as expected, the document was copied to dest, web-sink output was expected. Functions logs logged the `403` response, but only as a warning:

Screenshot 2024-03-27 at 16 30 29

### Web-sink off

With web sink off, we expect `ERRCONNREFUSED`, or similar error to be logged to functions logs and the function to fail. The function logs do show the error, after the main part of the function (the data trigger logic) had executed and report the function being killed due to unexpected error (all as expected):

Screenshot 2024-03-27 at 16 34 14

However, looking into the documents, it would appear that no rollback took place as the trigger did write the document as expected:

Screenshot 2024-03-27 at 16 32 58

ikusteu:
## Final setup: staging

I've created a setup so that the function can be deployed to production, allowing us to:
- specify where we wish the error to be thrown
- use custom sentry setup (if specified)

The function was deployed and all the behaviour observed was the same as one in local env. Error being thrown before/after sentry `flush` was reported by google cloud console and using a non-existing dsn, posing as Sentry server, did throw an error in an expected manner. However: **none of the errors thrown prevented the result documents from being written by the data trigger**

GC console error reports:
Screenshot 2024-03-29 at 21 23 04

Screenshot of data trigger results (despite the ENOTFOUND error):
Screenshot 2024-03-29 at 21 22 36

ikusteu:
With all of the tests performed, I don't see any relationship between Sentry error and data triggers producing db mismatches.

